Issue #8, opened 2019-11-15
add aliases for show unclaimed and show unspents


devhawk:
these `show` sub commands were named to match the underlying RPC endpoints. But this naming scheme is not intuitive for users.

* `show gas` should be an alias for `show unclaimed`
* `show unspent` should be an alias for `show unspents`

PR #24, opened 2019-12-02
show command user friendly output + aliases

### LINKED ISSUE #19: user friendly console output for show claimable / unspents / unclaimed
Instead of dumping JSON to the console, lets write this info in a human readable format. but add a --json option to retain the JSON representation for tooling scenarios

devhawk:
fixes #19 and #8

### DIFF
--- a/src/neo-express/Commands/ShowCommand.cs
+++ b/src/neo-express/Commands/ShowCommand.cs
@@ -1,4 +1,5 @@
 ﻿using McMaster.Extensions.CommandLineUtils;
+using Neo2Express.Models;
 using NeoExpress.Abstractions.Models;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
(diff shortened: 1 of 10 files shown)


djnicholson:
In order to not break the visual tracker (which depends on getting JSON output without passing a -json flag) we would need to break this into the following steps:

1. In neo-express allow a -json flag to be specified (but ignore it, JSON is always output anyway); release neo-express
2. In neo-visual-tracker start specifying the -json flag; release neo-visual-tracker
3. In neo-express start honoring the -json flag (and showing non-JSON by default); release neo-express

I'm also fine if you already considered this, but are choosing to temporarily break the (unreleased) tracker temporarily in order to simplify this change.  Just wanted to check it is an explicit decision and not an oversight :)

devhawk:
I'm ok w/ breaking devtracker master in the short term until you can add the --json argument. Opened https://github.com/neo-project/neo-visual-tracker/issues/29 to track on devtracker side

devhawk:
fixes #8

PR #41, opened 2020-02-24
Add GAS preload support

### LINKED ISSUE #40: Support GAS preload

devhawk:
This PR adds a --preload-gas support that will add enough blocks after the genesis block to give the genesis account the specified amount of gas. Note, the gas will be unavailable to start - the user will still needs to execute a transfer to release the GAS for claiming. 

Question for reviewers: Should we cap the amount of gas / blocks the user can specify? On my dev box, it takes about a second to add 200 blocks, which creates 1600 GAS. If we cap at 10 seconds (on my machine) that's around 2,000 blocks / 16,000 GAS

fixes #40

### DIFF
--- a/src/neo-express/Node/ExpressNodeRpcPlugin.cs
+++ b/src/neo-express/Node/ExpressNodeRpcPlugin.cs
@@ -313,7 +313,7 @@ namespace NeoExpress.Node
                 return JObject.Parse(json);
             }
 
-            throw new Exception("Unknown transaction");
+            return JObject.Null;
         }
 
         public JObject OnGetUnspents(JArray @params)
--- a/src/neo-express/Node/NodeUtility.cs
+++ b/src/neo-express/Node/NodeUtility.cs
@@ -149,12 +149,11 @@ namespace NeoExpress.Node
 
                         using (var snapshot = Blockchain.Singleton.GetSnapshot())
                         {
-                            var lastBlock = Blockchain.Singleton.GetBlock(snapshot.CurrentBlockHash);
-                            var validators = snapshot.GetValidators();
-                            if (lastBlock.Index == 0 && validators.Length == 1)
+                            if (PreloadValid())
                             {
-                                Plugin.Log("neo-express", LogLevel.Info, "Creating 1000 empty blocks to preload GAS");
-                                for (int i = 0; i < 125; i++)
+                                var preloadCount = 125;
+                                Plugin.Log("neo-express", LogLevel.Info, $"Creating {preloadCount} empty blocks to preload GAS");
+                                for (int i = 0; i < preloadCount; i++)
                                 {
                                     var block = CreatePreloadBlock(wallet);
                                     var reason = system.Blockchain.Ask<RelayResultReason>(block).Result;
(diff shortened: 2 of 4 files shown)


src/neo-express/Node/NodeUtility.cs
```diff
@@ -20,8 +23,123 @@ namespace NeoExpress.Node
 {
     internal static class NodeUtility
     {
-        public static Task RunAsync(Store store, ExpressConsensusNode node, TextWriter writer, CancellationToken cancellationToken)
+        private static ulong GetNonce()
         {
+            Span<byte> nonce = stackalloc byte[sizeof(ulong)];
+            Random rand = new Random();
```
djnicholson: a sequence of nonces might not be as random as you'd expect if this gets called in a loop (can make rand a static variable to avoid this problem).

djnicholson:
> Question for reviewers: Should we cap the amount of gas / blocks the user can specify? On my dev box, it takes about a second to add 200 blocks, which creates 1600 GAS. If we cap at 10 seconds (on my machine) that's around 2,000 blocks / 16,000 GAS

Probably not needed.  Maybe have it print a message to the console every 'n' blocks though (so it is obvious something is happening) and ensure that if the user does Ctrl+C they are still left in a runnable state (just with less GAS than they expected)?

devhawk:
> It would be good to claim the GAS too, so the user can immediately deploy a contract. I might be good to also create a wallet and transfer the GAS to that wallet (as users probably won't want to use the genesis wallet to deploy their contracts).

I'll take a look at adding a transfer and claim transaction. I explicitly *don't* want to create a wallet - part of my philosophy is that the developer should be in control of their setup rather than inheriting what we did. That's why I want to have the amount of gas preloaded be specified by the user rather than just an arbitrary value we decide on.

devhawk:
> > Question for reviewers: Should we cap the amount of gas / blocks the user can specify? On my dev box, it takes about a second to add 200 blocks, which creates 1600 GAS. If we cap at 10 seconds (on my machine) that's around 2,000 blocks / 16,000 GAS
> 
> Probably not needed. Maybe have it print a message to the console every 'n' blocks though (so it is obvious something is happening) and ensure that if the user does Ctrl+C they are still left in a runnable state (just with less GAS than they expected)?

Good suggestions

src/neo-express/Node/NodeUtility.cs
devhawk: Good catch. I fixed a similar issue for [Neo-FX master branch](https://github.com/neo-project/neo-fx/blob/master/src/NeoFx/StaticRandom.cs) but missed this one.

src/neo-express/Node/ExpressNodeRpcPlugin.cs
```diff
@@ -313,7 +313,7 @@ public JObject OnGetApplicationLog(JArray @params)
                 return JObject.Parse(json);
             }
 
-            throw new Exception("Unknown transaction");
+            return JObject.Null;
```
devhawk: @djnicholson this isn't an issue, right?

devhawk:
> I'll take a look at adding a transfer and claim transaction. 

Done.

djnicholson:
Opened https://github.com/neo-project/neo-visual-tracker/issues/119

PR #42, opened 2020-02-25
set UsePagerForHelpText before calling UseDefaultConventions

### LINKED ISSUE #39: When `neo-express run` and `neo-express checkpoint run` fail; don't block waiting for the user to hit 'q'
Right now it is hard to programmatically no whether neo-express started successfully not because the process stays running even when there is an error (it shows a help message and waits for a user to hit 'q')

devhawk:
fixes #39

### DIFF
--- a/src/neo-express/Program.cs
+++ b/src/neo-express/Program.cs
@@ -32,8 +32,10 @@ namespace NeoExpress
         {
             using (var app = new CommandLineApplication<Program>())
             {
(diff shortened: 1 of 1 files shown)


devhawk:
Note, I'm making this fix to the release/v1.0 branch. That means I will publish a new version of neo-express with this fix and I will re-merge this branch into master or cherry pick this fix over w/o review.

PR #43, opened 2020-02-25
Devhawk/fix37

### LINKED ISSUE #37: New design of express-get-populated-blocks requires O(n) RPC calls to render a single page of blocks in the dev tracker
We initially created the express-get-populated-blocks to avoid having to make O(n) RPC calls when rendering a filtered page of blocks in the dev tracker (where n is the size of the blockchain).

The new design re-introduces this characteristic.  It is 100x faster than it used to be (100 blocks can be checked at once) but still results in an unusable dev tracker UI experience when viewing a blockchain that has mostly unpopulated blocks (this will be common if a developer has left neo-express running generating empty blocks every 15s).

Suggest reverting https://github.com/neo-project/neo-express/pull/36. If we need a more efficient implementation of express-get-populated-blocks, we should maintain a persistent bitset of populated blocks within neo-express and update it as blocks are created.  The dev tracker cannot maintain this data structure as it changes in-between checkpoints.

devhawk:
This reverts commit d411befe6acc2ae4607dc0060e9e281013c839e0. Fixes #37 

I created this as a draft PR in order to discuss ramifications w/ @djnicholson. In particular, there are already versions of v1.0 w/ the current paging behavior out in the wild. I'm not sure if we should check this change into the release/v1.0 branch - meaning that some 1.0 installations will have the fix and some will not - or we should check into the master branch - meaning no one will get the fix until we ship the next version.

### DIFF
--- a/src/neo-express/Node/ExpressNodeRpcPlugin.cs
+++ b/src/neo-express/Node/ExpressNodeRpcPlugin.cs
@@ -428,30 +428,14 @@ namespace NeoExpress.Node
 
         private JObject OnGetPopulatedBlocks(JArray @params)
         {
-            using var snapshot = Blockchain.Singleton.GetSnapshot();
-
-            var count = @params.Count >= 1 ? uint.Parse(@params[0].AsString()) : 20;
-            count = count > 100 ? 100 : count;
-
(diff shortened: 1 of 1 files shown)


devhawk:
Way out of date. Closing for now. Will discuss w/ @djnicholson

PR #45, opened 2020-02-27
Preload on create, not run

### LINKED ISSUE #44: GAS preload should be a create command option
`--preload-gas` was added to the `run` command, but it makes much more sense to add it to the `create` command

devhawk:
move `--preload-gas` option from `run` command to `create` command.

fixes #44

### DIFF
--- a/src/neo-express/BlockchainOperations.cs
+++ b/src/neo-express/BlockchainOperations.cs
@@ -297,21 +297,24 @@ namespace NeoExpress
             }
         }
 
-        public static Task PreloadGasAsync(string directory, ExpressChain chain, int index, uint preloadGasAmount, TextWriter writer, CancellationToken cancellationToken)
+        public static void PreloadGas(string directory, ExpressChain chain, int index, uint preloadGasAmount, TextWriter writer, CancellationToken cancellationToken)
         {
             Debug.Assert(preloadGasAmount > 0);
-
-            chain.InitializeProtocolSettings(15);
+            if (!chain.InitializeProtocolSettings())
+            {
+                throw new Exception("could not initialize protocol settings");
+            }
             var node = chain.ConsensusNodes[index];
-#pragma warning disable IDE0067 // NodeUtility.PreloadAsync disposes the store when it's done
-            return NodeUtility.PreloadAsync(preloadGasAmount, new RocksDbStore(directory), node, writer, cancellationToken);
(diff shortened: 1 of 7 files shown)


src/neo-express/Node/NodeUtility.cs
```diff
@@ -172,23 +173,42 @@ void SubmitTransaction(Func<Snapshot, WalletAccount, Transaction?> factory)
...
             {
-                var generationAmount = Blockchain.GenerationAmount[0];
-                var gas = preloadGas / generationAmount;
-                return preloadGas % generationAmount == 0 ? gas : gas + 1;
+                if (i % 100 == 0)
+                {
+                    writer.WriteLine($"  Creating Block {i}");
+                }
+                var block = CreatePreloadBlock(wallet, random);
+                var relayResult = system.Blockchain.Ask<RelayResultReason>(block).Result;
```
vncoelho: Nice job, @devhawk.
Good workaround.

src/neo-express/BlockchainOperations.cs
```diff
@@ -296,26 +297,42 @@ private static void ValidateCheckpoint(string checkPointDirectory, long magic, E
             }
         }
 
-        public static Task RunBlockchainAsync(string directory, ExpressChain chain, int index, uint secondsPerBlock, uint preloadGas, TextWriter writer, CancellationToken cancellationToken)
+        public static void PreloadGas(string directory, ExpressChain chain, int index, uint preloadGasAmount, TextWriter writer, CancellationToken cancellationToken)
         {
-            chain.InitializeProtocolSettings(secondsPerBlock);
+            Debug.Assert(preloadGasAmount > 0);
```
djnicholson: it's a uint so this check probably isn't needed?

src/neo-express/Node/NodeUtility.cs
devhawk: If we create all the preload blocks on a single node, wouldn't those blocks automatically sync to the other nodes automatically? 

Filed #46 to track multi-node support

vncoelho: Yes, other nodes will sync automatically, @devhawk.

devhawk: THen it's probably just a matter of updating the signing code. Will do that as part of #46

PR #107, opened 2021-02-11
batch command

devhawk:
Batch mode enables Neo-express to perform a set of commands in a single process execution session. This can save considerable time relative to executing a single command per process session. For example, the registrar-sample executes a set of [seven commands](https://github.com/devhawk/registrar-sample/blob/4b3efd5f5fbedb945e470d66f2277532218ac9af/neo-express.targets#L3) during the build process. Without batch mode, a complete rebuild of registrar-sample takes just under 11 seconds (on my dev machine). With batch mode, a complete rebuild of registrar-sample only takes just over four seconds. 

A batch file is a simple text file with a separate express command on each line. Only the following commands are supported in batch files:

* contract deploy
* contract invoke
* checkpoint create
* oracle enable
* oracle response
* transfer


vncoelho (commented):
Looks very good and opens several possibilities. 
Great job!

JohndeVadoss (approved):

PR #119, opened 2021-03-23
add --witness-scope option to contract invoke + deploy

### LINKED ISSUE #115: Setting signer scopes
I tried to deploy a contract with `neoxp contract deploy /path/to/contract.nef SigningAccount`. The contract defines an owner address and in its `_deploy` method it calls `CheckWitness` to check if the singer of the deploy transaction is the owner. 
With the above command the `CheckWitness` call fails. The deployment transaction has a signer with witness scope `CalledByEntry`. When I do the deployment with other means and use the `Global` signer scope, the deployment transaction is successful. 

Is there a way to set the signer scope on neo-express commands like `contract deploy` or `contract invoke`?

devhawk:
fixes #115

cc @csmuller @gsmachado

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,6 +20,11 @@ jobs:
       run: dotnet restore
     - name: Pack 
       run: dotnet pack --output ./out --configuration Release --no-restore --verbosity normal
+    - name: Upload artifact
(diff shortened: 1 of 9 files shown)


JohndeVadoss (approved):

gsmachado:
awesome! 🎉

PR #137, opened 2021-05-05
Gracefully handle multiple contracts w/ same name in contract storage

### LINKED ISSUE #129: if two different wallets create a contract with the same name, neoxp contract storage ContractName fails with `An item with the same key has already been added. Key: ContractName`

devhawk:
fixes #129

### DIFF
--- a/src/neoxp/ExpressNodeExtensions.cs
+++ b/src/neoxp/ExpressNodeExtensions.cs
@@ -7,7 +7,6 @@ using System.Text;
 using System.Threading.Tasks;
 using Neo;
 using Neo.BlockchainToolkit;
-using Neo.BlockchainToolkit.Models;
 using Neo.Cryptography.ECC;
 using Neo.IO;
 using Neo.Network.P2P.Payloads;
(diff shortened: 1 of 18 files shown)


JohndeVadoss (approved):

src/neoxp/ExpressNodeExtensions.cs
```diff
@@ -45,7 +45,7 @@ public static async Task<ContractParameterParser> GetContractParameterParserAsyn
             };
 
             var contracts = await expressNode.ListContractsAsync().ConfigureAwait(false);
-            var lookup = contracts.ToDictionary(c => c.manifest.Name, c => c.hash);
+            var lookup = contracts.Distinct(new ContractNameEqualityComparer()).ToDictionary(c => c.manifest.Name, c => c.hash);
```
djnicholson, line 48: Assuming there are two contracts with the same name, does this cause one of them to arbitrarily ignored?  If so that might be a bad user experience.  I'd propose that we either block deployment of the second contract (so that it is not possible to reach this state) or be able to show the storage from each of the contracts.

I think it's quite likely that beginners will think that they update their contract code by redeploying it (I did!) so it's important to make it as easy as possible for them to realize that what they did didn't get them to the state that they wanted.

devhawk: Yes, this will remove the second (and any additional) contracts from the dictionary.  I'll take another look tomorrow at the possibility of blocking deployment - at least warning the user that they are deploying a 2nd (or additional!) contract with the same name

devhawk:
Significant changes since last review. @djnicholson take another look please

* `contract deploy` checks for name conflicts. `--force` option can be used to deploy contract with name conflict
* `contract get` and `contract storage` now handle specifying a contract name that matches multiple deployed contracts
* `tryGetContract` delegate in `GetContractParameterParserAsync` now throws if more than one contract matches provided name instead of simply taking the first value.

JohndeVadoss (approved):

djnicholson (approved):

PR #203, opened 2022-02-14
Generate Contract from keypair in DevWallet/DevWalletAccount

devhawk:
Instead of parsing contract details out of ExpressWalletAccount in FromExpressWalletAccount, use the private key to generate the signature contract. This ensures the contract is always up to date and not out of sync w/ the key.


devhawk:
Note, this PR is broken. Closing PR and taking another look

PR #209, opened 2022-03-15
Add command to download a contract 

ixje:
…from a remote network into our local chain of choice. This is part of an effort to make it easier to make our development (private) chain reflect the final network we want to deploy on. This is part 1, where part 2 will be on the `neo3-visual-tracker` repo. The end result could look like shown in the video.

https://user-images.githubusercontent.com/6625537/158369596-bca4fce7-a4e4-4c14-87db-0d3c7f963dd4.mp4

.


src/neoxp/Node/Modules/ExpressRpcMethods.cs
```diff
@@ -403,6 +405,90 @@ public JObject GetNep11Properties(JArray @params)
...
+            StorageItem item = snapshot.GetAndChange(key);
+            int value = (int)(BigInteger)item;
+            item.Add(1);
+            return value;
+        }
+        
+        [RpcMethod]
+        public JObject? ExpressPersistContract(JObject @params)
+        {
+            var state = @params[0]["state"];
```
ixje, line 438: I considered making it optional to override an existing contract in case the remote side updates the contract and we want to pull the new one. I believe the easier solution is deleting the existing contract (using point 1 of [this](https://github.com/ngdenterprise/neo3-visual-tracker/issues/133) visual tracker idea) and then downloading again. The downside is that it will also drop the storage which could require the user to call any functions that modified the storage again. But at least that should be obvious when the contract is deleted.

ixje:
The one thing I dislike about the current implementation is that the way of parameter entry to the download command is inconvenient. Selecting the destination chain is ok, then you can still have the contract hash you want to download on your clipboard for pasting, but the RPC url of the source chain (where we download the contract from) has to be typed manually. You can't seem to switch windows without losing the drop down box. 

I thought about having the option to store the "source rpc url" in some configuration file such that it can be selected from a list (or if the input field is empty takes the value from the config). What would be the right place? The `neo-servers.json` file perhaps?

src/neoxp/Commands/ContractCommand.cs
```diff
@@ -3,7 +3,7 @@
 namespace NeoExpress.Commands
 {
     [Command("contract", Description = "Manage smart contracts")]
-    [Subcommand(typeof(Deploy), typeof(Get), typeof(Hash), typeof(Invoke), typeof(List), typeof(Run), typeof(Storage))]
+    [Subcommand(typeof(Deploy), typeof(Get), typeof(Hash), typeof(Invoke), typeof(List), typeof(Run), typeof(Storage), typeof(Download))]
```
devhawk, line 6: nit: let's keep these in alphabetical order

src/neoxp/Commands/ContractCommand.Download.cs
```diff
@@ -0,0 +1,75 @@
...
+                {
+                    await writer.WriteLineAsync($"Invalid contract hash: {Contract} ").ConfigureAwait(false);
+                }
+                else
+                {
+                    // 1. Get ContractState
+                    var state = await this.rpcClient.GetContractStateAsync(Contract).ConfigureAwait(false);
+                    
+                    // 2. Get Full storage of the contract
+                    var storage_pairs = (JArray)await this.rpcClient.RpcSendAsync("getfullstorage", Contract).ConfigureAwait(false);
```
devhawk, line 53: I assume `getfullstorage` is a custom RPC endpoint? I'm not a fan of depending on an RPC method that isn't available as part of the base `RpcServer` module (except for ones exposed by Express itself obviously). Can we get `getfullstorage` added to `RpcServer`? 

Alternatively, we could use StateService RPC endpoints. Slightly more complex than `getfullstorage` but available by default on NGD provided RPC servers. The [`StateServiceStore`](https://github.com/ngdenterprise/neo-blockchaintoolkit-library/blob/master/src/bctklib/persistence/StateServiceStore.cs) used by `neo-trace` uses these endpoints. Would have the added benefit of being about to download contract + contract state at a specific blockchain height if desired.

src/neoxp/Commands/ContractCommand.Download.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                this.chainManagerFactory = chainManagerFactory;
+            }
+            
+            [Argument(0, Description = "Contract invocation hash")]
+            [Required]
+            internal string Contract { get; init; } = string.Empty;
+            
+            [Argument(1, Description = "Source network RPC address")]
+            internal string Source { get; init; } = string.Empty;
```
devhawk, line 31: The policy command also downloads content from a public chain. Policy command calls this argument `RpcUri`. We should be consistent. Also, the Policy command allows the user to specify "mainnet" or "testnet" in addition to supporting a full URL

src/neoxp/Commands/ContractCommand.Download.cs
```diff
@@ -0,0 +1,75 @@
...
+                }
+                else
+                {
+                    // 1. Get ContractState
+                    var state = await this.rpcClient.GetContractStateAsync(Contract).ConfigureAwait(false);
+                    
+                    // 2. Get Full storage of the contract
+                    var storage_pairs = (JArray)await this.rpcClient.RpcSendAsync("getfullstorage", Contract).ConfigureAwait(false);
+                    
+                    await writer.WriteLineAsync(storage_pairs.ToString(true)).ConfigureAwait(false);
```
devhawk, line 55: This should be argument controlled. Is the user usually going to want to see the raw storage data? I would think usually they would just want success confirmation when complete

src/neoxp/Commands/ContractCommand.Download.cs
```diff
@@ -0,0 +1,75 @@
...
+            [Required]
+            internal string Contract { get; init; } = string.Empty;
+            
+            [Argument(1, Description = "Source network RPC address")]
+            internal string Source { get; init; } = string.Empty;
+            
+            [Option(Description = "Path to neo-express data file")]
+            internal string Input { get; init; } = string.Empty;
+
+            internal async Task ExecuteAsync(TextWriter writer)
```
devhawk, line 36: Other commands break `ExecuteAsync` out into a separate method so it can be called as part of the `BatchCommand`. I'm thinking `download` should also be supported by batch.

src/neoxp/Node/OfflineNode.cs
```diff
@@ -406,5 +407,30 @@ IReadOnlyList<ExpressStorage> ListStorages(UInt160 scriptHash)
 
         public Task<IReadOnlyList<ExpressStorage>> ListStoragesAsync(UInt160 scriptHash)
             => MakeAsync(() => ListStorages(scriptHash));
+        
+        private const byte Prefix_Contract = 8;
+        
+        bool PersistContract(ContractState state, JArray storagePairs)
```
devhawk, line 413: The logic in this method appears twice - here and in OnlineNode. We should move the core downloaded contract persistence logic somewhere that can be shared by both

src/neoxp/Node/Modules/ExpressRpcMethods.cs
```diff
@@ -403,6 +405,90 @@ public JObject GetNep11Properties(JArray @params)
...
+            var c = new ContractState
+            {
+                Id = contractId,
+                Hash = UInt160.Parse(state["hash"].AsString()),
+                UpdateCounter = (ushort)state["updatecounter"].AsNumber(),
+                Nef = nef,
+                Manifest = m
+            };
+            
+            StorageKey key = new KeyBuilder(NativeContract.ContractManagement.Id, Prefix_Contract).Add(c.Hash);
```
devhawk, line 477: The logic in this method appears twice - here and in OfflineNode. We should move the core downloaded contract persistence logic somewhere that can be shared by both

src/neoxp/Node/Modules/ExpressRpcMethods.cs
```diff
@@ -403,6 +405,90 @@ public JObject GetNep11Properties(JArray @params)
...
+            {
+                contractId = GetNextAvailableId(snapshot);
+            }
+            else
+            {
+                // Update available id such that a regular contract deploy will use the right next id;
+                SetLastUsedContractId(snapshot, contractId);
+            }
+            
+            var c = new ContractState
```
devhawk, line 468: `RpcClient` has a `ContractStateFromJson` method. Why aren't you using that?

devhawk:
> The one thing I dislike about the current implementation is that the way of parameter entry to the download command is inconvenient. Selecting the destination chain is ok, then you can still have the contract hash you want to download on your clipboard for pasting, but the RPC url of the source chain (where we download the contract from) has to be typed manually. You can't seem to switch windows without losing the drop down box.

I left a comment about this in the code. TL;DR user should be able to specify 'mainnet' or 'testnet' in addition to a full URL.

> I thought about having the option to store the "source rpc url" in some configuration file such that it can be selected from a list (or if the input field is empty takes the value from the config). What would be the right place? The `neo-servers.json` file perhaps?

src/neoxp/Node/Modules/ExpressRpcMethods.cs
```diff
@@ -403,6 +405,90 @@ public JObject GetNep11Properties(JArray @params)
...
+                Script = Convert.FromBase64String(state["nef"]["script"].AsString())
+            };
+            
+            ContractManifest m = ContractManifest.FromJson(state["manifest"]);
+            
+            // Our local chain might already be using the contract id of the pulled contract, we need to check for this
+            // to avoid having contracts with duplicate id's. This is important because the contract id is part of the
+            // StorageContext used with Storage syscalls and else we'll potentially override storage keys or iterate
+            // over keys that shouldn't exist for one of the contracts.
+            int contractId = (int)state["id"].AsNumber();
```
devhawk, line 457: Shouldn't the offline implementation of contract download also be doing something like this? We need to consolidate the core logic for adding the downloaded contract/storage in one place so it's consistent between the two. This method should only be parsing the JSON and calling the shared function

src/neoxp/Node/Modules/ExpressRpcMethods.cs
devhawk: Would it make sense to make each optional? You know more about this scenario that I do, but it seems like each of these scenarios has value:

* Download remote contract/storage initially
* Update local contract storage from remote chain w/o updating contract
* Update local contract from remote chain w/o updating contract
* Re-download remote contract/storage, deleting existing copy + and contract state (though probably reuse contract id)

Note, whatever direction we go here the options need to be available on the download command itself, not just in the custom RPC method

src/neoxp/Node/Modules/ExpressRpcMethods.cs
devhawk: FYI, the `ExpressOracle` class provides static methods that are shared by on and offline node code. That might actually be the most logical place to put the common contract persistence logic, though we'd want to rename the class to something like `ExpressNodeCommon`

src/neoxp/Commands/ContractCommand.Download.cs
ixje: I understand your view and it's not ideal. I doubt we can get `getfullstorage` supported on the `RpcServer` because it potentially pulls a lot of storage (e.g. `NeoToken` storage is currently around 60 MB). Unless they're willing to accept it under the condition that we excluding native contracts.

Using `StateService` is not possible because it requires one to know which storage keys are used in the contract and not all contracts are open source. The work around is if monitor all contracts for `Storage.put/delete/get` syscalls in a 3rd party service, but that 3rd party service might as well run the RPC plugin that exposes `getfullstorage`. The `getfullstorage` is nothing more than this code
```C#
            byte[] prefix_key = StorageKey.CreateSearchPrefix(id, new byte[]{});

            JArray result = new JArray();
            foreach (var (key, value) in snapshot.Find(prefix_key))
            {
                JObject kv = new JObject();
                kv["k"] = Convert.ToBase64String(key.Key);
                kv["v"] = Convert.ToBase64String(value.Value);
                result.Add(kv);
            }

            return result;

```
I can give it a try to see what they think about it. I didn't do it because historically I've had much stronger use-cases for code additions and they've just been ignored.

src/neoxp/Commands/ContractCommand.cs
ixje: will fix

src/neoxp/Commands/ContractCommand.Download.cs
ixje: Good catch, shouldn't be there.

src/neoxp/Node/Modules/ExpressRpcMethods.cs
ixje: Didn't know that existed. I even created an issue https://github.com/neo-project/neo/issues/2674 to get it natively supported :sweat_smile:

src/neoxp/Commands/ContractCommand.Download.cs
devhawk: NeoTrace has the same requirement to download all the state for a contract w/o knowing the keys. The [`findstates`](https://docs.neo.org/docs/en-us/reference/rpc/latest-version/api/findstates.html) RPC method allows you to fetch all keys/value pairs for a given scriptHash at a given height. @ZhangTao1596 added this method for me specifically to enable NeoTrace.

src/neoxp/Node/Modules/ExpressRpcMethods.cs
devhawk: I would argue that all the from/to JSON methods should be in the core neo.dll not in modules, but since express already references RpcClient assembly anyway, no big deal.

src/neoxp/Commands/ContractCommand.Download.cs
ixje: Sorry I don't understand what's expected. I used `ContractCommand.Get` as base to implement the `Download` command. Looking at other `Contract` commands like `Deploy`, `Hash` and `Invoke` all I see is that they only implement `OnExecuteAsync` and not `ExecuteAsync`. The distinction between the 2 and their use-cases is not clear to me which is probably why I don't understand the request. If you can elaborate on this I can probably fix the code. Thanks

src/neoxp/Node/Modules/ExpressRpcMethods.cs
ixje: :100: there's quite a few places where the API in the core is asymmetrical exposing just with what they need, not what others might need.

src/neoxp/Node/Modules/ExpressRpcMethods.cs
ixje: want me to rename it as part of this PR?

ixje:
> > I thought about having the option to store the "source rpc url" in some configuration file such that it can be selected from a list (or if the input field is empty takes the value from the config). What would be the right place? The `neo-servers.json` file perhaps?
> 
> Express doesn't know anything about the `neo-servers.json` file. Not sure this command would get enough use to warrant adding that support

We can ignore this comment. It's actually for the `visual-tracker` side

src/neoxp/Commands/ContractCommand.Download.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                var (chainManager, _) = chainManagerFactory.LoadChain(Input);
+                var expressNode = chainManager.GetExpressNode();
+
+                if (!UInt160.TryParse(Contract, out var contractHash))
+                    throw new ArgumentException($"Invalid contract hash: \"{Contract}\"");
+
+                if (!TransactionExecutor.TryParseRpcUri(RpcUri, out var uri))
+                    throw new ArgumentException($"Invalid RpcUri value \"{RpcUri}\"");
```
ixje, line 44: Let me know if I should add `{}` around the throw. I noticed a mixed style for single line statements after a condition (with and without braces). This style came from one of the policy commands I copied the RpcUri parsing from.

devhawk: THanks for asking. I typically only skip braces when the entire if/then statement "looks good" on a single line. I'd appreciate if you'd add braces for these

src/neoxp/Commands/ContractCommand.Download.cs
```diff
@@ -0,0 +1,76 @@
...
+
+                using var rpcClient = new RpcClient(uri);
+                var stateAPI = new StateAPI(rpcClient);
+
+                var stateHeight = await stateAPI.GetStateHeightAsync();
+                if (stateHeight.localRootIndex is null)
+                    throw new Exception("Null \"localRootIndex\" in state height response");
+                var stateRoot = await stateAPI.GetStateRootAsync(stateHeight.localRootIndex.Value);
+                var states = await stateAPI.FindStatesAsync(stateRoot.RootHash, contractHash, new byte[0]);
```
devhawk, line 53: I don't think you're using this quite right. FindStates will return a configurable maximum of key/value pairs for each call (default 100). There's a `Truncated` property on the `RpcFoundStates` object that tells the caller if there are more records to retrieve. You can see a sync version of how to check `Truncated` here: https://github.com/ngdenterprise/neo-blockchaintoolkit-library/blob/master/src/bctklib/persistence/StateServiceStore.cs#L229

src/neoxp/Node/ExpressOracle.cs
```diff
@@ -138,5 +139,61 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+        private static int GetNextAvailableId(DataCache snapshot)
+        {
+            StorageKey key = new KeyBuilder(NativeContract.ContractManagement.Id, Prefix_NextAvailableId);
+            StorageItem item = snapshot.GetAndChange(key);
+            int value = (int)(BigInteger)item;
+            item.Add(1);
+            return value;
+        }
+        
+        public static bool PersistContract(SnapshotCache snapshot, ContractState state, (byte[] key, byte[] value)[] storagePairs)
```
devhawk, line 169: Yeah, I like this much better here. Please change the class/file name to `NodeUtility` or something similar

src/neoxp/Commands/ContractCommand.Download.cs
devhawk, line 53: Since you're going to have to update this code to handle `Truncated`...

I'm thinking it would make more sense *NOT* to decode the key/value pairs here. `FindStatesAsync` method decodes the storage pairs from base64 -> byte arrays and then `OnlineNode` turns around and re-encodes them in base64. Maybe it makes more sense to roll our own `FindStatesAsync` implementation that handles both `Truncated` and skips decoding storage items.  `IExpressNode.PersistContractAsync` signature would need to be updated and we could do Base64 -> byte[] decoding in `PersistContract`

Normally, I'm not a fan of leaking implementation details into an interface definition, but in this case, efficiency concerns might trump cleaner design.

src/neoxp/Node/ExpressOracle.cs
```diff
@@ -138,5 +139,61 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+            {
+                state.Id = GetNextAvailableId(snapshot);
+            }
+            else
+            {
+                // Update available id such that a regular contract deploy will use the right next id;
+                SetLastUsedContractId(snapshot, state.Id);
+            }
+            
+            StorageKey key = new KeyBuilder(NativeContract.ContractManagement.Id, Prefix_Contract).Add(state.Hash);
```
devhawk, line 185: What happens if the contract already exists? We need some type of `force` argument so the developer doesn't overwrite important data by accident

src/neoxp/Node/ExpressOracle.cs
```diff
@@ -138,5 +139,61 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+            
+            foreach (var pair in storagePairs)
+            {
+                snapshot.Add(
+                    new StorageKey { Id = state.Id, Key = pair.key}, 
+                    new StorageItem(pair.value)
+                );
+            }
+            snapshot.Commit();
+            return true;
```
devhawk, line 196: I'm thinking this method should return the contract ID that was used for the downloaded contract instead of a hardcoded true

src/neoxp/Node/Modules/ExpressRpcMethods.cs
```diff
@@ -3,18 +3,24 @@
 using System.Linq;
 using System.Numerics;
 using System.Threading;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
```
devhawk, line 6: ??? why are we using EF namespace here?

src/neoxp/Commands/ContractCommand.Download.cs
devhawk: My bad, I wasn't clear and my decription of the code was wrong.

Operations that modify the state of the blockchain are typically also supported by the `batch` command. Older versions of express would put the common code used by both direct and `batch` versions of the command into `ExecuteAsync`, but the `TransactionExecutor` class handles most of that now. Obviously, `download` modifications are not transactions, so `TransactionExecutor` would not be a good place to put `download` code shared between direct and `batch` versions.

I actually need to think about this more. THe only other command that connects to a public blockchain is `policy sync`, but the `batch` version of that command *explicitly does not support* downloading the policy from the public network when the `batch` command is run. I feel strongly that the `batch` command must be deterministic, so the `batch` version of `policy sync` only supports syncing with a previously downloaded policy state JSON file. By that logic, `batch` version of `download` should NOT support getting whatever the current version of contract + state is currently live on the public network. However, since we're using `StateService` anyway, we could force the `batch` version of `download` to specify a block height, ensuring deterministic update of the local express environment. 

Maybe it would be easier to discuss directly on Discord?

src/neoxp/Commands/ContractCommand.Download.cs
```diff
@@ -0,0 +1,76 @@
...
+                if (!UInt160.TryParse(Contract, out var contractHash))
+                    throw new ArgumentException($"Invalid contract hash: \"{Contract}\"");
+
+                if (!TransactionExecutor.TryParseRpcUri(RpcUri, out var uri))
+                    throw new ArgumentException($"Invalid RpcUri value \"{RpcUri}\"");
+
+                using var rpcClient = new RpcClient(uri);
+                var stateAPI = new StateAPI(rpcClient);
+
+                var stateHeight = await stateAPI.GetStateHeightAsync();
```
devhawk, line 49: Should also support specifying desired state height as an argument

src/neoxp/Commands/ContractCommand.Download.cs
devhawk: I pointed out above that the user should be able to specify block height for downloading contract + state, both for normal direct use and also for `batch` mode use. If the code knows the block height to download, we can move all the contract state download code into the `ExpressOracle.PersistContract` method. We wouldn't need to pass all the downloaded state to `ExpressPersistContract`, just the block height

src/neoxp/Node/Modules/ExpressRpcMethods.cs
ixje: No idea. I probably did some code in the past that required it and my IDE automatically added it. I've removed it together with other not used imports.

src/neoxp/Node/ExpressOracle.cs
ixje: The hardcoded `true` was just so I could use [MakeAsync()](https://github.com/neo-project/neo-express/blob/879d36a9a1548ad98a0a3bee6f6a389ab3840295/src/neoxp/Node/OfflineNode.cs#L94), `Id` sounds good to me.

src/neoxp/Node/ExpressOracle.cs
ixje: I think that's part of the discussion here https://github.com/neo-project/neo-express/pull/209#discussion_r826882935
As said on Discord I want to get the base in first and after that deal with the optional download parameters.

src/neoxp/Commands/ContractCommand.Download.cs
ixje: I was not happy with the decode, re-encode either but submitted as is to see what you found acceptable. 

As for moving everything to `ExpressOracle.PersistContract` it would end up requiring `RpcUri` and `block height` to be pushed down. If that's OK with you I'll make it happen. 

How about making `block height` optional and have it default to `-1` meaning get the latest state?

src/neoxp/Extensions/Extensions.cs
```diff
@@ -261,5 +262,48 @@ static bool TryGetNEP6Wallet(string path, string password, ProtocolSettings sett
                 }
             }
         }
+        
+        
+        private static JObject[] MakeFindStatesParams(
```
ixje, line 267: I'm aware this function also exists in `StateApi` except that version drops `count` (`pageSize`) if `from` is empty, this version doesn't.

src/neoxp/Commands/ContractCommand.Download.cs
devhawk: I'm fine with pushing `RpcUri` and `Height` down into `PersistContract`, but I think the calling code needs to pass a valid block height. For direct mode, if the user doesn't pass a specific height, the command execution code can retrieve the current height to pass to `PersistContract`. For batch mode, the batch height is required.

src/neoxp/Node/ExpressOracle.cs
devhawk: I'm fine getting the base line code working and approved and then dealing with handling overwriting. Neo platform is planning to ship [v3.2 release](https://github.com/neo-project/neo/issues/2676) in a couple weeks. As long as we get `contract download` in before that, I don't think we're in a rush otherwise

src/neoxp/Commands/ContractCommand.Download.cs
```diff
@@ -0,0 +1,81 @@
...
+                if (stateHeight.localRootIndex is null)
+                {
+                    throw new Exception("Null \"localRootIndex\" in state height response");
+                }
+
+                var stateRoot = await stateAPI.GetStateRootAsync(stateHeight.localRootIndex.Value);
+                var states = await rpcClient.ExpressFindStatesAsync(stateRoot.RootHash, contractHash, new byte[0]);
+                var contractState = await rpcClient.GetContractStateAsync(Contract).ConfigureAwait(false);
+
+                await expressNode.PersistContractAsync(contractState, states.Results);
```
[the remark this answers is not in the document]
devhawk: No, that doesn't seem like information the user can use for anything

devhawk:
> Btw do you prefer 1 commit per issue or are you fine with the 1 commit for all I've been doing so far?

I don't typically review commit by commit. I like looking at all the changes at once. So whatever works better for you

src/neoxp/Extensions/Extensions.cs
devhawk: Ugh, you're right. Any chance you can open a PR on neo modules to fix it there too? That way we'll eventually be able to remove the express copy of this code.

src/neoxp/Commands/ContractCommand.Download.cs
devhawk: Yup, but you'll also need to add a `--height` command line option and only retrieve the current height if `--height` isn't specified.

BTW, I just noticed that some commands with value type options use a nullable type (`StopCommand.NodeIndex`) while others use an explicit `hasValue` tuple (`OracleCommand.RequestId`). Please use the explicit pattern. At some point, I'll clean up the versions that use nullable type so express is consistent

src/neoxp/Node/ExpressOracle.cs
ixje: We can still do it in this PR. Just doing it last of all

src/neoxp/Commands/ContractCommand.Download.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.IO;
+using System.Threading.Tasks;
+using McMaster.Extensions.CommandLineUtils;
+using NeoExpress.Node;
+
+namespace NeoExpress.Commands
+{
+    partial class ContractCommand
+    {
+        internal enum ContractForce
```
devhawk, line 12: Nit: call this OverwriteForce

src/neoxp/Extensions/Extensions.cs
devhawk: I filed https://github.com/neo-project/neo-express/issues/219 to remind myself to remove the local copy when the updated one in neo-modules is available

src/neoxp/IExpressNode.cs
```diff
@@ -40,6 +42,7 @@ enum CheckpointMode { Online, Offline }
         Task<IReadOnlyList<ExpressStorage>> ListStoragesAsync(UInt160 scriptHash);
         Task<IReadOnlyList<TokenContract>> ListTokenContractsAsync();
 
+        Task<int> PersistContractAsync(ContractState state, (string key, string value)[] storagePairs, ContractCommand.ContractForce? force);
```
devhawk, line 45: I'm not a fan of nullable enums. Can we add "None" as the default force parameter?

src/neoxp/Node/OfflineNode.cs
```diff
@@ -406,6 +408,14 @@ IReadOnlyList<ExpressStorage> ListStorages(UInt160 scriptHash)
 
         public Task<IReadOnlyList<ExpressStorage>> ListStoragesAsync(UInt160 scriptHash)
             => MakeAsync(() => ListStorages(scriptHash));
+        
+        int PersistContract(ContractState state, (string key, string value)[] storagePairs, ContractCommand.ContractForce? force)
```
devhawk, line 412: There's no need to have the PersistContract method since it's just a one liner. Just put the call to NodeUtility.PersistContract inside the MakeAsync call of PersistContractAsync

src/neoxp/Node/NodeUtility.cs
```diff
@@ -138,5 +142,127 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+            {
+                throw new ArgumentException($"Invalid contract hash: \"{contractHash}\"");    
+            }
+
+            if (!TransactionExecutor.TryParseRpcUri(rpcUri, out var uri))
+            {
+                throw new ArgumentException($"Invalid RpcUri value \"{rpcUri}\"");
+            }
+
+            if (isBatchCommand && stateHeight == 0)
```
devhawk, line 188: We should check for stateHeight != 0 in batch command and remove isBatchCommand param

src/neoxp/Node/NodeUtility.cs
```diff
@@ -138,5 +142,127 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+            return value;
+        }
+
+        public static async Task<(ContractState contractState, (string key, string value)[] storagePairs)> ProcessDownloadParamsAsync(
+            string contractHash,
+            string rpcUri,
+            uint stateHeight,
+            bool isBatchCommand)
+        {
+            if (!UInt160.TryParse(contractHash, out var contractHash_))
```
devhawk, line 178: nit: C# more commonly has the underscore at the front of the variable

src/neoxp/Extensions/ExpressRpcFoundStates.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Neo.IO.Json;
+
+namespace NeoExpress
+{
+    public class ExpressRpcFoundStates
```
devhawk, line 8: nit: move to Models folder / namespace

src/neoxp/Extensions/Extensions.cs
```diff
@@ -266,5 +267,48 @@ static bool TryGetNEP6Wallet(string path, string password, ProtocolSettings sett
...
+            jobjectArray[2] = (JObject) Convert.ToBase64String(prefix);
+            jobjectArray[3] = (JObject) Convert.ToBase64String(from);
+            if (count.HasValue)
+                jobjectArray[4] = (JObject) (double) count.Value;
+            return jobjectArray;
+        }
+        public static async Task<ExpressRpcFoundStates> ExpressFindStatesAsync(this RpcClient rpcClient, UInt256 rootHash, UInt160 contractScriptHash, ReadOnlyMemory<byte> prefix, ReadOnlyMemory<byte> from = default, int? pageSize = null)
+        {
+            ExpressRpcFoundStates? states = null;
+            var start = from.ToArray();
```
devhawk, line 291: start param should be a `ReadOnlyMemory<byte>` and we can avoid converting the from param to an array.

src/neoxp/Extensions/ExpressRpcFoundStates.cs
```diff
@@ -0,0 +1,34 @@
...
+                        j["key"].AsString(),
+                        j["value"].AsString()
+                    ))
+                    .ToArray(),
+                FirstProof = ProofFromJson(json["firstProof"]),
+                LastProof = ProofFromJson(json["lastProof"]),
+            };
+        }
+
+        private static (bool hasValue, byte[]? value) ProofFromJson(JObject json)
```
devhawk, line 31: param should be `JObject? json` if json can be null

src/neoxp/Extensions/ExpressRpcFoundStates.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Linq;
+using Neo.IO.Json;
+
+namespace NeoExpress
+{
+    public class ExpressRpcFoundStates
+    {
+        public bool Truncated;
+        public (string key, string value)[] Results = new (string key, string value)[0];
+        public (bool hasValue, byte[]? value) FirstProof;
```
devhawk, line 12: Why do we have both `hasValue` and a nullable `value`? Seems like one or the other would suffice.

src/neoxp/Extensions/ExpressRpcFoundStates.cs
```diff
@@ -0,0 +1,34 @@
...
+                        j["value"].AsString()
+                    ))
+                    .ToArray(),
+                FirstProof = ProofFromJson(json["firstProof"]),
+                LastProof = ProofFromJson(json["lastProof"]),
+            };
+        }
+
+        private static (bool hasValue, byte[]? value) ProofFromJson(JObject json)
+            => json == null ? (false, null) : (true, Convert.FromBase64String(json.AsString()));
```
devhawk, line 32: Probably can leave the proofs encoded too, since we never look at them

src/neoxp/Extensions/ExpressRpcFoundStates.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Neo.IO.Json;
+
+namespace NeoExpress
+{
+    public class ExpressRpcFoundStates
+    {
+        public bool Truncated;
+        public (string key, string value)[] Results = new (string key, string value)[0];
```
devhawk, line 11: `Array.Empty<(string, string)>()` avoids an allocation

src/neoxp/Extensions/Extensions.cs
```diff
@@ -266,5 +267,48 @@ static bool TryGetNEP6Wallet(string path, string password, ProtocolSettings sett
...
+            while (true)
+            {
+                var @params = MakeFindStatesParams(rootHash, contractScriptHash, prefix.Span, start, pageSize);
+                var response = await rpcClient.RpcSendAsync("findstates", @params).ConfigureAwait(false);
+                var foundStates = ExpressRpcFoundStates.FromJson(response);
```
devhawk, line 296: Might make more sense to process the JSON here directly rather than deserializing into a class like ExpressRpcFoundStates. We only need `truncated` and `results` from `response.

src/neoxp/Extensions/Extensions.cs
```diff
@@ -266,5 +267,48 @@ static bool TryGetNEP6Wallet(string path, string password, ProtocolSettings sett
...
+            JObject[] jobjectArray = new JObject[count.HasValue ? 5 : 4];
+            jobjectArray[0] = (JObject) rootHash.ToString();
+            jobjectArray[1] = (JObject) scriptHash.ToString();
+            jobjectArray[2] = (JObject) Convert.ToBase64String(prefix);
+            jobjectArray[3] = (JObject) Convert.ToBase64String(from);
+            if (count.HasValue)
+                jobjectArray[4] = (JObject) (double) count.Value;
+            return jobjectArray;
+        }
+        public static async Task<ExpressRpcFoundStates> ExpressFindStatesAsync(this RpcClient rpcClient, UInt256 rootHash, UInt160 contractScriptHash, ReadOnlyMemory<byte> prefix, ReadOnlyMemory<byte> from = default, int? pageSize = null)
```
devhawk, line 288: We only use the `Results` field from `ExpressRpcFoundStates` so lets just return that. Also, if we return a `Task<IEnumerable<(string key, string value)>>` we can avoid interim `ToArray` calls - ExpressFindStatesAsync can simply concat the results of each `findstates` call together an the caller can call `ToArray` on the return value

src/neoxp/Extensions/Extensions.cs
```diff
@@ -266,5 +267,48 @@ static bool TryGetNEP6Wallet(string path, string password, ProtocolSettings sett
...
+                var @params = MakeFindStatesParams(rootHash, contractScriptHash, prefix.Span, start, pageSize);
+                var response = await rpcClient.RpcSendAsync("findstates", @params).ConfigureAwait(false);
+                var foundStates = ExpressRpcFoundStates.FromJson(response);
+                
+                if (states is null)
+                {
+                    states = foundStates;
+                } else
+                {
+                    states.Results = states.Results.Concat(foundStates.Results).ToArray();
```
devhawk, line 303: see comment above about removing the interum ToArray call

src/neoxp/Node/NodeUtility.cs
```diff
@@ -138,5 +142,127 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+        private static int GetNextAvailableId(DataCache snapshot)
+        {
+            StorageKey key = new KeyBuilder(NativeContract.ContractManagement.Id, Prefix_NextAvailableId);
+            StorageItem item = snapshot.GetAndChange(key);
+            int value = (int)(BigInteger)item;
+            item.Add(1);
+            return value;
+        }
+
+        public static async Task<(ContractState contractState, (string key, string value)[] storagePairs)> ProcessDownloadParamsAsync(
```
devhawk, line 172: Call this `DownloadContractAsync`

src/neoxp/Node/NodeUtility.cs
```diff
@@ -138,5 +142,127 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+            {
+                if (force is null)
+                {
+                    throw new Exception("Contract already exists locally. Use --force:<option> to overwrite");
+                }
+                if (force == ContractCommand.ContractForce.All ||
+                    force == ContractCommand.ContractForce.ContractOnly)
+                {
+                    // Note: a ManagementContract.Update() will not change the contract hash. Not even if the NEF changed.
+                    localContract.Nef = state.Nef;
```
devhawk, line 248: I don't think setting localContract fields automatically saves the data back the storage. You're only modifying the in-memory copy you got from calling `GetContract`

src/neoxp/Node/NodeUtility.cs
```diff
@@ -138,5 +142,127 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+                    localContract.UpdateCounter = state.UpdateCounter;
+                }
+            }
+
+            if (force == ContractCommand.ContractForce.All || force == ContractCommand.ContractForce.StorageOnly)
+            {
+                foreach (var pair in storagePairs)
+                {
+                    snapshot.Add(
+                        new StorageKey { Id = state.Id, Key = Convert.FromBase64String(pair.key) },
```
devhawk, line 259: What happens if `localContract.Id != state.Id`?

src/neoxp/Commands/BatchCommand.cs
```diff
@@ -122,6 +123,20 @@ internal async Task ExecuteAsync(IDirectoryInfo root, ReadOnlyMemory<string> com
                                 cmd.Model.Force).ConfigureAwait(false);
                             break;
                         }
+                    case CommandLineApplication<BatchFileCommands.Contract.Download> cmd:
+                       {
+                        var expressNode = chainManager.GetExpressNode();
```
devhawk, line 128: nit: indentation

devhawk: Also, txExec already an expressNode instance, so you shouldn't call `GetExpressNode` here. we should probably a wrapper `ContractDownloadAsync` method to `TransactionExecutor`

src/neoxp/IExpressNode.cs
ixje: I thought about that initially then assumed `nullable` things were the more standard C# approach. Will fix

src/neoxp/Extensions/Extensions.cs
ixje: it cannot be readonly because we update `start` on line `310`

src/neoxp/Node/OfflineNode.cs
ixje: I thought we always wanted to have a `sync` version as well, but happy to remove

src/neoxp/Extensions/ExpressRpcFoundStates.cs
ixje: `json` cannot be `null`?

src/neoxp/Extensions/Extensions.cs
ixje: I like the idea, but I can't get this to work. Something about the function body that can't be an iterator block due to the return type not being an iterator interface type (probably due to `Task<>`). Can you do this change instead?

src/neoxp/Node/NodeUtility.cs
ixje: oof that was sloppy from my side

src/neoxp/Node/NodeUtility.cs
ixje: I didn't think that scenario was feasible. For that to occur the following needs to happen
1. deploy contract on remote network
2. deploy the same contract on the private network using the same wallet as used on the remote network
3. download remote contract to private network

I thought using the same wallet is already highly unlikely and then why would you pull the remote contract again. Either way, I added a check where we just re-use the existing `localContract.id`.

src/neoxp/Commands/BatchCommand.cs
ixje: You previously (https://github.com/neo-project/neo-express/pull/209#discussion_r830643990) said
> Obviously, download modifications are not transactions, so TransactionExecutor would not be a good place to put download code shared between direct and batch versions.

Now that you mentioned `nodes` I actually thought of 2 possible issues
1. I think the current code currently only works on a single node chain
2. Clients connected to our consensus node won't know the storage changed because it is not done through a transaction in a block that is relayed over the network :thinking: 

   This might actually be hard to solve. First thing that comes to mind is creating a transaction script that does all the modifications, but only the [contract itself](https://github.com/neo-project/neo/blob/2a64c1cc809d1ff4b3a573c7c22bffbbf69a738b/src/neo/SmartContract/ApplicationEngine.Storage.cs#L79) can modify it's own storage otherwise we'd have a security issue. I have to think more about this

src/neoxp/Extensions/Extensions.cs
devhawk: If we make `start` a `ReadOnlyMemory<byte>` type, we can't modify the contents of `start`, but we can still re-assign the `start` variable to a different `ReadOnlyMemory<byte>` instance (i.e. the return value of `Convert.FromBase64String`). 

Unlike say JavaScript or TypeScript, C# doesn't really have the concept of const or read only local variables. If you specify a const, it has to be a compile time constant, not something you assign at runtime. 

I can see why this is confusing to the non C# dev

src/neoxp/Extensions/ExpressRpcFoundStates.cs
devhawk: Null has been called the ["billion dollar mistake"](https://www.lucidchart.com/techblog/2015/08/31/the-worst-mistake-of-computer-science/). C# 8 and beyond attempts to undo this mistake by allowing developers to opt-in to declaring if their variables are expected to be null or not. That's what the question mark after the type name means - `JObject? json` means `json` could be `null`, `JObject json` means that `json` should not be null.

Note, the underlying runtime still allows for null variable instances, the nullable checking is done via flow analysis not runtime enforcement. But I like to be consistent and I've opted all managed neo projects that I manage into tracking nullability this way.

src/neoxp/Extensions/Extensions.cs
devhawk: Will do

src/neoxp/Node/NodeUtility.cs
devhawk: Then we should be explicitly checking for that scenario and throwing

src/neoxp/Commands/BatchCommand.cs
devhawk: I'm OK if this feature only works on single node chains. We just need to add a check at the beginning to make sure that we're on a single node chain and throwing if we're not.

As far as I can tell, there's no way we'll be able to replicate the updated storage to other  non-consensus nodes connected to the blockchain. I'm not sure how common that scenario is tho.

devhawk: Also <sigh> throwing my own words back to me... 

We don't want to create another instance of expressNode. So we can either ignore my previous advice *or* we can create a public property on TransactionExecutor to be able to get the express node instance. Your call

devhawk:
FYI, I updated `ExpressFindStatesAsync`.  In addition to my comment earlier about being able to reassign the `start` variable, I also want to clarify that IEnumerable is processed lazily. In other words, if I have an `IEnumerable<(string key, string value)>` instance, the items in the enumeration don't get evaluated until you actually for each over the IEnumerable value. Calling `ToArray` does that, but there's no need to do that until you've retrieved all the values from the public chain. If I call .ToArray multiple times, I'm just doing that lazy evaluation over and over for no reason. 

I also changed the code to pass storage pairs around as a read only list instead of an array. Calling .ToList is slightly more efficient than ToArray, and we just loop over the pairs when persisting anyway so no need to use array

src/neoxp/Node/NodeUtility.cs
ixje: It's not invalid, just highly unlikely. It's solved now, so now it's just a "good catch!" on your side 👍

src/neoxp/Commands/BatchCommand.cs
ixje: 1. Regarding single vs multi-node support. Can't we just call download and persist contract in a loop for all consensus nodes? Worst case we need to stop the nodes for a moment to update and then restart them. I don't know if batch mode is actually limited to single node?

2. I think we might be able to solve the chain replication using a [P2P plugin](
https://github.com/neo-project/neo/blob/2a64c1cc809d1ff4b3a573c7c22bffbbf69a738b/src/neo/Plugins/IP2PPlugin.cs#L26). We can put the contractstate + storage in a custom payload, wrap that in an [ExtensiblePayload](https://github.com/neo-project/neo/blob/2a64c1cc809d1ff4b3a573c7c22bffbbf69a738b/src/neo/Network/P2P/Payloads/ExtensiblePayload.cs). You can then still run a normal client, it just needs the plugin to handle the payload.

   I think we should handle this in a separate PR.

devhawk: BTW, `TransactionExecutor` already has a `ExpressNode` property. `checkpoint create` and `fastfwd` commands in batch mode both use the `ExpressNode` property already. `contract download` can use it to.

We could totally look at doing something like download and persist in a loop. That might work for checkpoints, which are currently limited to single node chains. I'll think about that and the P2P plugin idea some more today.

src/neoxp/Node/NodeUtility.cs
```diff
@@ -138,5 +142,127 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+                {
+                    snapshot.Delete(k);
+                }
+
+                for (int i = 0; i < storagePairs.Count; i++)
+                {
+                    snapshot.Add(
+                        new StorageKey { Id = contractId, Key = Convert.FromBase64String(storagePairs[i].key) },
+                        new StorageItem(Convert.FromBase64String(storagePairs[i].value)));
+                }
```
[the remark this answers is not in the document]
devhawk: It's been my understanding that for loops are slightly more performant than foreach for indexed types. foreach creates an enumerator object while indexed access does not. But it's been a long time since I learned that. I will do some research on the current best practices. Let's leave it for loop for now to be consistent with the rest of the code base.

src/neoxp/Node/NodeUtility.cs
```diff
@@ -138,5 +142,127 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+        {
+            StorageKey key = new KeyBuilder(NativeContract.ContractManagement.Id, Prefix_NextAvailableId);
+            StorageItem item = snapshot.TryGet(key);
+            return (int)(BigInteger)item;
+        }
+
+        private static void SetLastUsedContractId(DataCache snapshot, int newId)
+        {
+            StorageKey key = new KeyBuilder(NativeContract.ContractManagement.Id, Prefix_NextAvailableId);
+            StorageItem item = snapshot.GetAndChange(key);
```
devhawk, line 159: I think item needs to be incremented by one here. You're calling `SetLastUsedContractId` with the id of the contract downloaded from remote chain. Pretty sure the next contract deployed will overwrite the downloaded contract

src/neoxp/Node/NodeUtility.cs
devhawk: Have we validated that setting the value returned from GetInteroperable also updates the raw value? I don't think it does

src/neoxp/Node/NodeUtility.cs
ixje: This is not needed because a contract deploy will add +1 to the last used id, see [deploy](https://github.com/neo-project/neo/blob/2a64c1cc809d1ff4b3a573c7c22bffbbf69a738b/src/neo/SmartContract/Native/ContractManagement.cs#L187) and [GetNextAvailableId](
https://github.com/neo-project/neo/blob/2a64c1cc809d1ff4b3a573c7c22bffbbf69a738b/src/neo/SmartContract/Native/ContractManagement.cs#L79-L83)

src/neoxp/Extensions/Extensions.cs
ixje: Mmmh I swear I remember my IDE warning me, but I believe what you're saying. Seems solved now anyway :)

src/neoxp/Node/NodeUtility.cs
ixje: I can see where the confusion might come from; `Prefix_NextAvailableId` is a poor name choice (I kept it the same as the `neo` source). It's actually more a `Prefix_LastUsedId`

devhawk: > I can see where the confusion might come from; `Prefix_NextAvailableId` is a poor name choice (I kept it the same as the `neo` source). It's actually more a `Prefix_LastUsedId`

I'm not sure that's right. My read of this is that the stored value is returned as the ID to use while storage is updated to increment by one. If the currently stored value is 10, GetNextAvailableId returns 10 and updates storage to be 11.

``` csharp
        private int GetNextAvailableId(DataCache snapshot)
        {
            StorageItem item = snapshot.GetAndChange(CreateStorageKey(Prefix_NextAvailableId));
            int value = (int)(BigInteger)item;
            item.Add(1);
            return value;
        }
```

src/neoxp/Commands/ContractCommand.Download.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                var (chainManager, _) = chainManagerFactory.LoadChain(Input);
+                var expressNode = chainManager.GetExpressNode();
+
+                var result = await NodeUtility.DownloadParamsAsync(Contract, RpcUri, Height);
+                await expressNode.PersistContractAsync(result.contractState, result.storagePairs, Force);
+            }
+            
+            internal async Task<int> OnExecuteAsync(CommandLineApplication app, IConsole console)
+            {
```
devhawk, line 58: Did we decide that we need to limit contract download to single node express chains?

src/neoxp/Commands/BatchCommand.cs
```diff
@@ -122,6 +123,24 @@ internal async Task ExecuteAsync(IDirectoryInfo root, ReadOnlyMemory<string> com
...
                         }
+                    case CommandLineApplication<BatchFileCommands.Contract.Download> cmd:
+                        {
+                            if (cmd.Model.Height == 0)
+                            {
+                                throw new ArgumentException("Height cannot be 0. Please specify a height > 0");
+                            }
+                            
+                            var expressNode = chainManager.GetExpressNode();
+                            var result = await NodeUtility.DownloadParamsAsync(
```
devhawk, line 134: Might make sense to create a static helper method on `ContractCommand.Download` to do validation (single node chain, height > 0) and the calls to `DownloadParamsAsync` and `PersistContractAsync`. Not critical, but might be good to have validation all in one place

src/neoxp/Commands/ContractCommand.Download.cs
ixje: for this PR yes. Then in a new one we can fix for multiple consensus + chain replication

src/neoxp/Commands/BatchCommand.cs
ixje: Making sure I understand the request right; is the concern that the above 2 lines are duplicated in both `BatchCommand` and `ContractCommand.Download.ExecuteAsync()`? 
fwiw; only `BatchCommand` has the `height > 0` validation to ensure deterministic behaviour. `ContractCommand` on the other hand has `height = 0` as default where it means query the chain for the latest height.

src/neoxp/Node/NodeUtility.cs
```diff
@@ -138,5 +142,127 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+        private static int GetNextAvailableId(DataCache snapshot)
+        {
+            StorageKey key = new KeyBuilder(NativeContract.ContractManagement.Id, Prefix_NextAvailableId);
+            StorageItem item = snapshot.GetAndChange(key);
+            int value = (int)(BigInteger)item;
+            item.Add(1);
+            return value;
+        }
+
+        public static async Task<(ContractState contractState, IReadOnlyList<(string key, string value)> storagePairs)> DownloadParamsAsync(
```
devhawk, line 172: Can we change the name of this? `DownloadParamsAsync` isn't really clear what params it's doing. Maybe `DownloadContractStateAsync` (yes, I know it's also downloading storage but I think just saying state is fine)

src/neoxp/Commands/BatchCommand.cs
devhawk: "Concern" is a bit of an overstatement. Introducing a method to just abstract away the calls to `DownloadParamsAsync` and `PersistContractAsync` isn't worth it. I was thinking that if you also validated `height > 0` and `consensus node count == 1`, it might become worth it. 

To do this, you'd need to lift the `stateAPI.GetStateHeightAsync` call out of `DownloadParamsAsync` and only call it in the direct call implementation (i.e. not from the batch command). However, I hadn't thought thru the ramifications of that change. Lifting `GetStateHeightAsync` out would then also force you to lift the `TransactionExecutor.TryParseRpcUri` call out since you need the URL to make `GetStateHeightAsync` call. At that point, we would have lost the value of the abstraction anyway.

Please do this:
* Add a check in `PersistContractAsync` implementation to ensure consensus node count is 1
* Also add a check before calling `DownloadParamsAsync`. No point of downloading the contract data only to throw when we go to persist
* Note, I left a separate comment that `DownloadParamsAsync` needs a better name

ixje: I followed the request. Note that points 1 and 2 add quite some duplication, was that the intention? What we get now as flow is
- check node count
- call DownloadContractStateAsync
- call PersistContractAsync
    (internally)
   - check node count
   - call NodeUtility.PersistContract

Perhaps it's not worth spending too much time on as we'll quickly switch to multi-node support after this PR is merged and remove those checks again.

devhawk: It's a minor change, but I think there should be node count check in `NodeUtility.PersistContract` - utility methods like that need to make sure the operation they are about to make are allowed in the current context. Any other check we add would be for optimization - bailing out early before doing any work if the command is not allowed. 

* We should check node count as soon as possible after recognizing the user issued the contract download command. The checks that you added to BatchCommand and ContractCommand.Download are exactly what I wanted. 
* I can see the value in checking node count in `OnlineNode.PersistContractAsync`. Arguably better to bail out before making the RPC call
* The node check in `OfflineNode.PersistContractAsync` should simply be moved to `NodeUtility.PersistContract`. No need to check in both places.

src/neoxp/Node/NodeUtility.cs
```diff
@@ -208,6 +211,131 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+                    force == ContractCommand.OverwriteForce.ContractOnly)
+                {
+                    // Note: a ManagementContract.Update() will not change the contract hash. Not even if the NEF changed.
+                    localContract.Nef = state.Nef;
+                    localContract.Manifest = state.Manifest;
+                    localContract.UpdateCounter = state.UpdateCounter;
+                }
+            }
+
+            if (force == ContractCommand.OverwriteForce.All || force == ContractCommand.OverwriteForce.StorageOnly)
```
devhawk, line 317: if force is none and the contract didn't previously exist, the storage records won't be persisted

src/neoxp/Node/NodeUtility.cs
```diff
@@ -208,6 +211,131 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+                {
+                    // Update available id such that a regular contract deploy will use the right next id;
+                    SetNextAvailableContractId(snapshot, state.Id + 1);
+                }
+
+                snapshot.Add(key, new StorageItem(state));
+            }
+            else
+            {
+                if (force == ContractCommand.OverwriteForce.None)
```
devhawk, line 302: We should check to see if the contract value matches the downloaded value, and only throw if there would be a need to overwrite

src/neoxp/Node/NodeUtility.cs
```diff
@@ -208,6 +211,131 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+            var contractState = await rpcClient.GetContractStateAsync(contractHash).ConfigureAwait(false);
+
+            return (contractState, states);
+        }
+
+        public static int PersistContract(SnapshotCache snapshot, ContractState state,
+            IReadOnlyList<(string key, string value)> storagePairs, ContractCommand.OverwriteForce force)
+        {
+            StorageKey key = new KeyBuilder(NativeContract.ContractManagement.Id, Prefix_Contract).Add(state.Hash);
+            var localContract = snapshot.GetAndChange(key)?.GetInteroperable<ContractState>();
```
devhawk, line 280: since the value returned from `GetAndChange` was passed to `GetInteroperable` and then discarded, I don't think updates to localContract will be reflected in storage

devhawk: I take it back, it appears to work just fine.

src/neoxp/Node/NodeUtility.cs
devhawk: I have validated this

src/neoxp/Node/NodeUtility.cs
ixje: We can validate `local contract state == downloaded contract state` and the same for all storage pairs, then give more specific errors ("Contract already exists - state and storage differs", ".. - storage differs", ".. - state differs" or "Contract already exists - aborting"). That would be friendlier to the user.

src/neoxp/Node/NodeUtility.cs
```diff
@@ -208,6 +212,186 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+                if (force == ContractCommand.OverwriteForce.None)
+                {
+                    IList<(string key, string value)> states = new Collection<(string key, string value)>();
+                    byte[] prefixKey = StorageKey.CreateSearchPrefix(localContract.Id, new byte[]{});
+                    foreach (var (k, v) in snapshot.Find(prefixKey))
+                    {
+                        states.Add((Convert.ToBase64String(k.Key.ToArray()), Convert.ToBase64String(v.ToArray())));
+                    }
+                    
+                    var stateEquals = localContract.ToJson().ToByteArray(false)
```
ixje, line 339: There is probably a better way of doing this, but at least it works (where `Equals` and `==` don't).

src/neoxp/Node/NodeUtility.cs
```diff
@@ -208,6 +212,186 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+                        .SequenceEqual(state.ToJson().ToByteArray(false));
```
ixje, line 340: There is probably a better way of doing this, but at least it works (where `Equals` and `==` don't).

src/neoxp/Node/NodeUtility.cs
```diff
@@ -208,6 +211,186 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+            StorageKey key = new KeyBuilder(NativeContract.ContractManagement.Id, Prefix_Contract).Add(state.Hash);
+            var localContract = snapshot.GetAndChange(key)?.GetInteroperable<ContractState>();
+
+            if (localContract is null)
+            {
+                // Our local chain might already be using the contract id of the pulled contract, we need to check for this
+                // to avoid having contracts with duplicate id's. This is important because the contract id is part of the
+                // StorageContext used with Storage syscalls and else we'll potentially override storage keys or iterate
+                // over keys that shouldn't exist for one of the contracts.
+                if (state.Id <= LastUsedContractId(snapshot))
```
devhawk, line 315: Nit: Why not just always use the next available ID? Since we have to handle the case where the local id != remote id, why not simplify the logic?

src/neoxp/Node/NodeUtility.cs
```diff
@@ -208,6 +211,197 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+        {
+            var (overwriteContract, overwriteStorage) = force switch 
+            {
+                ContractCommand.OverwriteForce.All => (true, true),
+                ContractCommand.OverwriteForce.ContractOnly => (true, false),
+                ContractCommand.OverwriteForce.None => (false, false),
+                ContractCommand.OverwriteForce.StorageOnly => (false, false),
+                _ => throw new NotSupportedException($"Invalid OverwriteForce value {force}"),
+            };
+
```
devhawk, line 309: Note, I pulled GetSnapshot into PersistContract so that the snapshot lifetime can be managed with a using statement

src/neoxp/Node/NodeUtility.cs
ixje: I tried to keep it the same as the remote contract whenever possible. Always using the last is also an option.

src/neoxp/Node/NodeUtility.cs
```diff
@@ -208,6 +211,197 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+        public static int PersistContract(NeoSystem neoSystem, ContractState state,
+            IReadOnlyList<(string key, string value)> storagePairs, ContractCommand.OverwriteForce force)
+        {
+            var (overwriteContract, overwriteStorage) = force switch 
+            {
+                ContractCommand.OverwriteForce.All => (true, true),
+                ContractCommand.OverwriteForce.ContractOnly => (true, false),
+                ContractCommand.OverwriteForce.None => (false, false),
+                ContractCommand.OverwriteForce.StorageOnly => (false, false),
```
ixje, line 306:
```suggestion
                ContractCommand.OverwriteForce.StorageOnly => (false, true),
```

src/neoxp/Node/NodeUtility.cs
```diff
@@ -208,6 +211,197 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+                    localContract.UpdateCounter = state.UpdateCounter;
+                }
+                else
+                {
+                    // check to make sure the existing contract matches the downloaded contract
+                    // except for id, which is allowed to be different from downloaded contract
+                    if (!localContract.Hash.Equals(state.Hash)
+                        || localContract.UpdateCounter != state.UpdateCounter
+                        || !localContract.Nef.ToArray().SequenceEqual(state.Nef.ToArray())
+                        || !localContract.Manifest.ToJson().ToByteArray(false).SequenceEqual(state.Manifest.ToJson().ToByteArray(false)))
```
ixje, line 348: This part feels strange to have. If we reach this point then the following conditions hold
* We found a contract locally with the same `hash`  as `state.hash` (the remote contract) at L313. Therefore 
   ```C#
   !localContract.Hash.Equals(state.Hash)
   ```
   is always `false`
* the `force` argument is `StorageOnly` || `NONE`
   * if `force` = `StorageOnly` we do nothing because the logic between lines 356-399 deals with storage
   * if `force` = `NONE` then we're not supposed to change anything and should always fail. 

I think the only logic that should be here is the detailed explanation what part of the contract already exists (as I thought was requested here https://github.com/neo-project/neo-express/pull/209#discussion_r845574402) but those changes are now gone and we actually get a generic error instead.

src/neoxp/Node/NodeUtility.cs
```diff
@@ -208,6 +211,197 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+
+                for (int i = 0; i < storagePairs.Count; i++)
+                {
+                    snapshot.Add(
+                        new StorageKey { Id = contractId, Key = Convert.FromBase64String(storagePairs[i].key) },
+                        new StorageItem(Convert.FromBase64String(storagePairs[i].value)));
+                }
+            }
+            else
+            {
```
ixje, line 373: If the downloaded contract is new and `force` is `NONE`. It will fail on the `storageCount` mismatch throwing a misleading error that storage already exists while there's actually nothing. There's also no logic to persist the storage of a completely new contract.

src/neoxp/Node/NodeUtility.cs
```diff
@@ -208,6 +211,197 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+            // the ContractState as it was at stateHeight. ContractManagement stores ContractState by
+            // contractHash with the prefix 8. The following code uses stateAPI.GetStateAsync to retrieve
+            // the value with that key at the height state root and then deserializes it into a ContractState
+            // instance via GetInteroperable.
+
+            var key = new byte[21];
+            key[0] = 8; // ContractManagement.Prefix_Contract
+            _contractHash.ToArray().CopyTo(key, 1);
+
+            var contractStateBuffer = await stateAPI.GetStateAsync(
```
ixje, line 290: throws 
```C#
Neo.Network.RPC.RpcException: The given key was not present in the dictionary.
```
if the remote network does not have the request contract

ixje:
I tried downloading the `PolicyContract` which already exists but got no error, which seems incorrect.

src/neoxp/Node/NodeUtility.cs
devhawk: Good catch. I have a solution for this in the sync version of GetState I wrote for bctk library: https://github.com/ngdenterprise/neo-blockchaintoolkit-library/blob/master/src/bctklib/persistence/RpcClientExtensions.cs#L35

src/neoxp/Node/NodeUtility.cs
devhawk: Ouch, bad copy/paste error

src/neoxp/Node/NodeUtility.cs
devhawk: I think it's simpler just to issue a new local contract id

src/neoxp/Node/NodeUtility.cs
devhawk: WIll fix

src/neoxp/Node/NodeUtility.cs
```diff
@@ -208,6 +211,197 @@ public static void SignOracleResponseTransaction(ProtocolSettings settings, Expr
...
+                stateRoot.RootHash, NativeContract.ContractManagement.Hash, key);
+            var contractState = new StorageItem(contractStateBuffer).GetInteroperable<ContractState>();
+            var states = await rpcClient.ExpressFindStatesAsync(stateRoot.RootHash, _contractHash, default);
+
+            return (contractState, states);
+        }
+
+        public static int PersistContract(NeoSystem neoSystem, ContractState state,
```
devhawk, line 298: How should we handle the user trying to download a native contract? Obviously, we can't support downloading native contract state, but we could support downloading native contract storage. Not sure it's a good idea tho.

ixje:
As an end user you have no idea what the error means. I think we should catch it and throw something along the lines of “contract not found on remote server”.

src/neoxp/Node/NodeUtility.cs
devhawk: I will catch and throw an exception that makes more sense to the user

JohndeVadoss (approved):

PR #244, opened 2022-08-15
Add optional data parameter to transfer

### LINKED ISSUE #198: Add optional data parameter to transfer
As part of #197, we should also add an optional `--data` argument to transfer command that will be passed as the `data` parameter of the [NEP17 transfer operation](https://github.com/neo-project/proposals/blob/master/nep-17.mediawiki#transfer). Today, the data parameter is hard coded to null

devhawk:
Fixes #198

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -17,13 +17,25 @@ may not exactly match a publicly released version.
 
 ### NeoExpress
 
-### Changed
+#### Added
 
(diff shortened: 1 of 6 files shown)


devhawk:
FYI, merging w/o approval as @JohndeVadoss is OOF

Issue #290, opened 2023-11-02
Inappropriate command description documentation


chenzhitong:
```
Usage: neoxp create [options] <Output>

Arguments:
  Output                                  name of .neo-express file to create (Default: ./default.neo-express)

Options:
  -c|--count <COUNT>                      Number of consensus nodes to create
                                          Default: 1
                                          Allowed values are: 1, 4, 7.
  -a|--address-version <ADDRESS_VERSION>  Version to use for addresses in this blockchain instance
                                          Default: 53
  -f|--force                              Overwrite existing data
```

In this case, based on the information provided in the documentation, the "Output" parameter can be considered as an optional parameter because it has a clear default value. Therefore, you can use square brackets **[ ]** to indicate that it is optional. Here is the updated command description:

```
Usage: neoxp create [options] [Output]

Arguments:
  Output                               name of .neo-express file to create (Default: ./default.neo-express)
```
This will more accurately reflect the nature of the "Output" parameter, which means users can choose to provide a custom file name or opt to omit it and use the default file name.

PR #291, opened 2023-11-02
Update command-reference.md

chenzhitong:
Fix https://github.com/neo-project/neo-express/issues/290

### DIFF
--- a/docs/command-reference.md
+++ b/docs/command-reference.md
@@ -51,18 +51,17 @@ can be specified in multiple ways:
 ## neoxp create
 
 ```
-Usage: neoxp create [options] <Output>
+Usage: neoxp create [Options] [Output]
 
 Arguments:
-  Output                                  name of .neo-express file to create (Default: ./default.neo-express)
-
-Options:
+[Options]:
   -c|--count <COUNT>                      Number of consensus nodes to create
                                           Default: 1
                                           Allowed values are: 1, 4, 7.
   -a|--address-version <ADDRESS_VERSION>  Version to use for addresses in this blockchain instance
                                           Default: 53
   -f|--force                              Overwrite existing data
+[Output]: name of .neo-express file to create (Default: ./default.neo-express)
 ```
 
 The `create` command is used to create a new Neo-Express blockchain network for local development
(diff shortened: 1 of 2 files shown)


lock9:
Hi @chenzhitong ,
I've noticed you updated the installation instructions. These will work but it's not how it should be used. Neo-express _requires no installation_. It can, and should, be shipped as a standalone application. I don't think that asking people to install the .NET SDK is reasonable. They are not going to develop anything, they don't need to install the SDK. We can embed the runtime with the application.

chenzhitong:
> Hi @chenzhitong , I've noticed you updated the installation instructions. These will work but it's not how it should be used. Neo-express _requires no installation_. It can, and should, be shipped as a standalone application. I don't think that asking people to install the .NET SDK is reasonable. We can embed the runtime in the application.

I copied the installation guide directly from the readme.md. Isn't neo-express installed through dotnet tools? And the installation guide doesn't say anything about installing the SDK.

lock9:
>Isn't neo-express installed through dotnet tools?
>And the installation instructions doesn't say anything about installing the SDK.

`dotnet tools` is part of the .NET SDK. We can continue to distribute it using it. However, neo-express can be compiled and shipped as a standalone app that doesn't require external tools (like dotnet tools). 
It may be fine to update the installation guide now, but be aware that we need to change the distribution method.

chenzhitong:
I'll revert the installation instructions first and then submit it after neo-express is compiled into a standalone application

chenzhitong:
I'll revert the installation instructions first and then submit it after neo-express is compiled into a standalone application

chenzhitong:
I'll revert the installation instructions first and then submit it after neo-express is compiled into a standalone application @lock9

docs/command-reference.md
```diff
@@ -714,20 +699,19 @@ local file, can be used as the input for the `policy sync` command described bel
...
-  Account                   Account to pay contract invocation GAS fee
-
-Options:
+[Options]:
   -p|--password <PASSWORD>  password to use for NEP-2/NEP-6 sender
   -i|--input <INPUT>        Path to neo-express data file
   -t|--trace                Enable contract execution tracing
   -j|--json                 Output as JSON
+<Policy>: Policy to set
+                            Allowed values are: GasPerBlock, MinimumDeploymentFee, CandidateRegistrationFee,
```
lock9, line 711: There are some differences in indentation.

docs/command-reference.md
```diff
@@ -736,19 +720,18 @@ The `policy set` command updates the current value of the specified Neo-Express
...
-  Account                   Account to pay contract invocation GAS fee
-
-Options:
+[Options]:
   -p|--password <PASSWORD>  password to use for NEP-2/NEP-6 sender
   -i|--input <INPUT>        Path to neo-express data file
   -t|--trace                Enable contract execution tracing
   -j|--json                 Output as JSON
+<Source>: Source of policy values. Must be local policy settings JSON file or the URL of Neo JSON-RPC
+                            Node
```
lock9, line 732: There are some differences in indentation.

lock9 (commented):
I saw some differences in the indentation, but everything else looks fine. By the way, the next release has an 'update' command that needs to be described here. Let's leave that for later.

lock9 (approved):

lock9:
Hi @Liaojinghui, can you merge this PR? Thanks

Liaojinghui (approved):

Issue #274, opened 2023-03-31
Automatically parse String responses


lock9:
Currently, when a method with a return type of "String" is called in Neo Express, the returned value is displayed in hexadecimal format. For instance, the "symbol" method returns "4c494e4b44" instead of the expected value "LINKD".

To fix this, I suggest adding a feature that automatically converts string return types to UTF-8 before displaying them. This would make it easier for developers to read and interpret the output of methods that return strings.

For instance, the "symbol" method in the contract's manifest specifies that it returns a string, as shown in the code snippet below:
```json
{
    "name": "symbol",
    "offset": 0,
    "parameters": [],
    "returntype": "String",
    "safe": true
}
```

However, the current output of the method is in hexadecimal format, which can be difficult to read and interpret:
<img width="154" alt="Screenshot 2023-03-31 at 10 11 31" src="https://user-images.githubusercontent.com/38396062/229129176-1dbd52b9-28cb-4024-92e6-dc3b0e468d3a.png">

I suggest that neo-express (and related tools) convert hex to UTF-8, when the method return type is String

chenzhitong:
If the return value is 2222, then what will be displayed, display "" or \"\"? 
If the former, it will make the json report an error, if the latter, it will lose the meaning of the conversion.
I suggest this tool for conversion https://neo.org/converter/index

lock9:
If the method return type is a string, can't we convert it to utf-8?
We don't need the extra quotes. What if we print the result between line breaks?

cschuchardt88:
Its just almost all things in the `vm` are stored as bytes. It was just a lazy way to display generic data, instead of parsing or validating it. Not checking the contract method return type is the problem.

cschuchardt88:
@lock9 
I believe #316 does fix this issue? If so can we close?

PR #316, opened 2023-11-09
optimize invoke output

Ashuaidehao:
Try to convert ByteString to string when invoke contract with "--result".
Partial solution of https://github.com/neo-project/neo-express/issues/274.


cschuchardt88 (commented):
we should be checking the manifest for the return type to see if it is `string`

Ashuaidehao:
Hardly check manifest outside of the vm :( .

src/neoxp/Extensions/Extensions.cs
```diff
@@ -57,6 +58,28 @@ public static void WriteWarning(string value)
             Console.WriteLine($"\x1b[33mWarning: {value}\x1b[0m");
         }
 
+        public static bool TryGetUtf8String(this ReadOnlySpan<byte> data, out string text)
+        {
+            try
+            {
+                text = Encoding.UTF8.GetString(data);
```
cschuchardt88, line 65: Wouldn't be easier to use Strict UTF-8?

src/neoxp/Extensions/Extensions.cs
```diff
@@ -57,6 +58,28 @@ public static void WriteWarning(string value)
...
+                byte[] reencodedBytes = Encoding.UTF8.GetBytes(text);
+                for (int i = 0; i < data.Length; i++)
+                {
+                    if (data[i] != reencodedBytes[i])
```
cschuchardt88, line 69: Could check for  `char.IsAscii()`

src/neoxp/TransactionExecutor.cs
```diff
@@ -278,7 +278,14 @@ static async Task WriteStackItemAsync(System.IO.TextWriter writer, Neo.VM.Types.
                         await WriteLineAsync(Neo.Helper.ToHexString(buffer.GetSpan())).ConfigureAwait(false);
                         break;
                     case Neo.VM.Types.ByteString byteString:
-                        await WriteLineAsync(Neo.Helper.ToHexString(byteString.GetSpan())).ConfigureAwait(false);
+                        if (byteString.GetSpan().TryGetUtf8String(out var text))
+                        {
+                            await WriteLineAsync($"{Neo.Helper.ToHexString(byteString.GetSpan())}({text})").ConfigureAwait(false);
```
cschuchardt88, line 283: I think text 1st then hex. Or put on different lines when printing.

Example: `Return: "Hello World" (Hex-Format: "48656c6c6f20576f726c64")'

src/neoxp/Extensions/Extensions.cs
Ashuaidehao: It will get very low perfomance to catch the exception when using random `data` with Strict UTF-8. `try-catch` here is use for robust, and I don't expect it runs into `catch` case.

src/neoxp/Extensions/Extensions.cs
Ashuaidehao: I'm sorry I don't get it. If the data is encoded utf-8 string, checking `IsAscii` is not enough.

src/neoxp/TransactionExecutor.cs
Ashuaidehao: It is not a accurate convert, some random data could be converted to string. So I prefer show its text as a possible value after the certain value(HexString).

Liaojinghui:
Merge?

src/neoxp/TransactionExecutor.cs
cschuchardt88: Yes but you are showing string and hex-string. Think it neater for it to look like this `"Hello World" (Hex-Format: "48656c6c6f20576f726c64")` and more understandable.

src/neoxp/Extensions/Extensions.cs
cschuchardt88: You allow strings to be displayed that are not human readable.

cschuchardt88:
@Liaojinghui i still think that we should be checking the manifest for return type string. The current way is dirty and inefficient.

src/neoxp/TransactionExecutor.cs
cschuchardt88: You have the `StackItem`, why not use `StackItem.GetString()`? that will fix this problem.

src/neoxp/Extensions/Extensions.cs
Ashuaidehao: If utf-8 convert successfully, it is already readable without extra checking.

src/neoxp/TransactionExecutor.cs
Ashuaidehao: `StackItem.GetString()` will throw exception if it is not a valid string.

lock9:
Hello!

I'm a bit worried about this PR 🤔 It doesn't look like I expected it to be 🤔 🤔 🤔 

Can we do a post-processing step to parse the JSON and use the manifest to ensure String compatibility?

src/neoxp/Extensions/Extensions.cs
cschuchardt88: ok thought you were using ASCII for some reason... 😂

Ashuaidehao:
> Hello!
> 
> I'm a bit worried about this PR 🤔 It doesn't look like I expected it to be 🤔 🤔 🤔
> 
> Can we do a post-processing step to parse the JSON and use the manifest to ensure String compatibility? @Ashuaidehao

Yes, but the best way to implement it requires neo-vm expose more infomations, it may cost more time to discuss and couldn't process custom scripts without contract invoking. This PR is just a quick way show possible value to developers, and the developers should know what the return value is.

cschuchardt88:
> the developers should know what the return value is.

What if result comes from an oracle?

Ashuaidehao:
> > the developers should know what the return value is.
> 
> What if result comes from an oracle?

Could oracle results show here? I'm not sure if  I miss some scenarios.

cschuchardt88:
I dont remember, but i can't see why not.

cschuchardt88 (approved):

Liaojinghui (approved):

PR #329, opened 2023-11-12
Revert "Inconsistent account naming"

cschuchardt88:
Reverts neo-project/neo-express#305

This is a big change. This breaks old batch files that have been developed by users.

Sorry 😢 

Ill see if i can get some unit tests going for `neo-express` as well to prevent things like this.


Liaojinghui:
I would say making mistake is ok, we can fix it.

Liaojinghui (approved):

PR #335, opened 2023-11-15
Use testnet/mainnet network configuration

### LINKED ISSUE #332: Use testnet/mainnet network configuration
Problem: Today, users run Neo Express using fees different from those on Mainnet and Testnet. This makes the development environment different from the one used by developers. It's common for developers to sync their fees using the `sync` command.

Solution options:
- Hardcode mainnet and testnet values
- Use the `sync` command during setup

cschuchardt88:
Closes #332 

**Change Log**
- Added sync policy on create command

image

### DIFF
--- a/src/neoxp/Commands/CreateCommand.cs
+++ b/src/neoxp/Commands/CreateCommand.cs
@@ -10,8 +10,10 @@
 // modifications are permitted.
 
 using McMaster.Extensions.CommandLineUtils;
+using Neo.BlockchainToolkit;
 using Neo.Network.RPC;
-using NeoExpress.Models;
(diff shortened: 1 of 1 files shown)


lock9 (changes requested):
UT? 😅 
Some inline comments could also help

cschuchardt88:
UT would be nice, But were to start... I test everything by debugging, Yes it isn't good enough. That's all we have atm. I'm waiting to see if `Blockchain-toolkit-library` get added to this repo... so i can rebuild and get the UTs

Liaojinghui (approved):

PR #336, opened 2023-11-15
Import WIF

### LINKED ISSUE #297: Automatically create a user account during 'setup'
Problem: 
Neo Express uses a multi-sig account as the 'starting account'. This prevents users from importing the private key on other devices (they expect single sig accounts). 

Proposed Solution:
This happens because the genesis account is responsible for running the consensus algorithm. Replacing it for a single sig account is not possible. We should create a new account and transfer part of the NEO and GAS to this new account. We can call it 'dev'.

cschuchardt88:
Hopefully closes #297 now

**Change Log**
- Added WIF suuport for importing private key

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ jobs:
       uses: dotnet/nbgv@v0.4.0
       id: nbgv
     - name: Install dependencies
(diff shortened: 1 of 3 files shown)


Liaojinghui (approved):

cschuchardt88:
fixed now

PR #347, opened 2023-11-16
Added wallet create to batch

### LINKED ISSUE #297: Automatically create a user account during 'setup'

cschuchardt88:
Closes #297 - now completely

**Change Log**
- Added wallet create command to batch files
- Changed formatting and text for wallet create command

**Example `setup.batch`**
```
wallet create bob
transfer 10000 gas genesis bob
```

### DIFF
--- a/src/neoxp/Commands/BatchCommand.BatchFileCommands.cs
+++ b/src/neoxp/Commands/BatchCommand.BatchFileCommands.cs
@@ -19,7 +19,7 @@ namespace NeoExpress.Commands
     partial class BatchCommand
     {
         [Command]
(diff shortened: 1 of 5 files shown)


Liaojinghui (approved):

Liaojinghui:
I suggest to add code security scan action

cschuchardt88:
github action?

PR #388, opened 2023-11-26
Bypass publish

cschuchardt88:


.github/workflows/push.yml
```diff
@@ -22,3 +22,6 @@ jobs:
   pack:
     needs: [test]
     uses: ./.github/workflows/pack.yml
+  publish:
+    needs: [pack]
+    uses: ./.github/workflows/pack.yml
```
Liaojinghui, line 28: Why pack and publish has the same workflow

cschuchardt88:
@Liaojinghui merge please, let see what happens now

Liaojinghui (approved):

.github/workflows/push.yml
cschuchardt88: to see it works with publish the secret

PR #389, opened 2023-11-26
Removed publish from workflow

cschuchardt88:
added back `linux-musl-arm64` as well, we need it.


Liaojinghui (approved):

Liaojinghui:
@cschuchardt88  fail on push

PR #393, opened 2023-11-27
avoid uploading artifacts

Liaojinghui:
this pr focus on deleting all logics that upload artifacts to the github.


Liaojinghui:
close as wrong branch base

PR #406, opened 2023-11-28
Fix code.yml

cschuchardt88:
We have cron job that runs. So we won't be checking for changes in code files. Or else it won't run. If that is the case we can remove cron job.


Liaojinghui:
What i mean is we do not check the file if they are not updated.

cschuchardt88:
but there is a cron job that runs though. If i add that logic cron job won't work. Now we can remove cron job.

Issue #362, opened 2023-11-18
Add Standalone Packages to...


cschuchardt88:
We should open the options for `neo-express` with soon to be standalone packages with #358.
The reason for doing this will allow users to use the app without installing dotnet. Also random searches in these repos may bring new users to the platform after seeing the `neo-express` in action.

### Repos

**Windows Package Manager (winget)**
Setup and [Submit packages](https://learn.microsoft.com/en-us/windows/package-manager/package/). After you create the initial request. I can setup `github` action workflow to deploy new releases.

**apt-get**
Can use github repo for this [Setup](https://wiki.debian.org/DebianRepository/SetupWithReprepro).

**dotnet tool**
we already use this method.

**neo.org/downloads.html**
Have a download page for `neo-express`. This would be a big benefit seeing how no one knows about this app. Let alone where or how to download it.

**brew**
I don't know how

@shargon @Liaojinghui

I would like to know if any of this is possible with the `neo-project`. What the process of getting all this setup? Cause we will need a new hosting repo for `apt-get`, I'm pretty sure `brew` and any others. Maybe `nccs`, `neo-cli` will follow in suite?

Liaojinghui:
This should be handled by NGD. @steven1227 @Ashuaidehao @superboyiii

chenzhitong:
If a standalone packages is created, I can add a link to it here https://neo.org/neogas#wallets

cschuchardt88:
@chenzhitong #358 is in the process of getting approved for images.

cschuchardt88:
well #358 is now live. And word on the future for images on neo.org @chenzhitong ?

I would really like to add to `winget`. It would have to be this repo to do the initial PR to `winget` to get the package started

chenzhitong:
Can you write a profile of neo-express, describing what kind of wallet it is.
@cschuchardt88 
And what's the download page, GitHub Release? There's no release file to download yet!

cschuchardt88:
@Liaojinghui when is a release coming out? so we can get a download. 

@chenzhitong its not a wallet, its a developer node, put it here https://neo.org/dev#tooling

chenzhitong:
We need a logo for neo-express

cschuchardt88:
![neo-express-icon](https://github.com/neo-project/neo-express/assets/8141309/9705e7fa-7d06-44f5-a7e9-2720652b2a76)

cschuchardt88:
[...]

PR #408, opened 2023-12-01
Fix the problem that double-click exe program can't be run

chenzhitong:
After releasing as a standalone application, beginners might run it by double-clicking directly rather than starting it from the terminal. In such cases, provide a prompt when no parameters are passed.


cschuchardt88:
If we get #362 rolling. This will be a small issue

chenzhitong:
> If we get #362 rolling. This will be a small issue

Doesn't this Issue discuss how to publish and download programs? Does this issue discuss involve program interaction?

Jim8y:
@cschuchardt88

cschuchardt88:
@Jim8y These will affect a lot of other tools. So I would say no. Unless @chenzhitong has a good reason.

Jim8y:
This will not be added

PR #412, opened 2023-12-05
Debian workflow

cschuchardt88:
This adds two new packages to the releases.


Liaojinghui (approved):

Liaojinghui:
How would those envs being set?

cschuchardt88:
being used in the current standalone linux environment

Liaojinghui:
saw it

PR #415, opened 2023-12-06
Workflow fixes again 🤣 🍭 

cschuchardt88:
Fixed workflow again. The changes from @Liaojinghui were not running.

Everything tested


Liaojinghui:
So we are not checking the file type anymore?

cschuchardt88:
I have it only works on `*cs` and `*csproj` for different workflow files. See `pr.yml` and `pr_cs.yml`

PR #468, opened 2024-11-27
Modify the document link

chenzhitong:
After some discussion, the documentation was changed back to docs.neo.org


Jim8y:
changed back means we have to keep maintaiing doc repo right? it is out of date

cschuchardt88 (approved):

chenzhitong:
> changed back means we have to keep maintaiing doc repo right? it is out of date

Right now both docs.neo.org and developers.neo.org are using the https://github.com/neo-project/neo-dev-portal repository for their data sources, so there's no need to change that. docs.neo.org is based on the `dev` branch, and developers.neo.org is based on the `master` branch.

PR #503, opened 2025-10-07
Update packages

shargon:
Solved npm issue:

> xml2js versions before 0.5.0 allows an external attacker to edit or add new properties to an object. This is possible because the application does not properly validate incoming JSON keys, thus allowing the __proto__ property to be edited.

> The latest possible version that can be installed is 0.4.23 because of the following conflicting dependencies:
vsce@2.15.0 requires xml2js@^0.4.23
No patched version available for xml2js
The earliest fixed version is 0.5.0.

image

### DIFF
--- a/src/bctklib/bctklib.csproj
+++ b/src/bctklib/bctklib.csproj
@@ -14,7 +14,7 @@
         <InternalsVisibleTo Include="test.bctklib" />
     </ItemGroup>
     <ItemGroup>
-        <PackageReference Include="MessagePack" Version="3.1.4" />
+        <PackageReference Include="MessagePack" Version="2.5.198" />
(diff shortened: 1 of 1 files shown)


shargon:
@ajara87 could you help me with the 4 ut that fault?

src/bctklib/bctklib.csproj
```diff
@@ -14,17 +14,17 @@
         <InternalsVisibleTo Include="test.bctklib" />
     </ItemGroup>
     <ItemGroup>
-        <PackageReference Include="MessagePack" Version="2.5.198" />
+        <PackageReference Include="MessagePack" Version="3.1.4" />
```
shargon, line 17:
```suggestion
        <PackageReference Include="MessagePack" Version="2.5.198" />
```

PR #519, opened 2026-02-04
Add appengine provider create

ajara87:


src/neoxp/Node/Modules/ExpressApplicationEngineProvider.cs
```diff
@@ -35,7 +35,22 @@ class ExpressApplicationEngineProvider : IApplicationEngineProvider
 
             return null;
         }
+        public ApplicationEngine Create(TriggerType trigger, IVerifiable container, DataCache snapshot, Block persistingBlock, ProtocolSettings settings, long gas, IDiagnostic diagnostic, JumpTable jumpTable)
```
shargon: Why this method?

PR #540, opened 2026-04-25
Fix Quick Start checkpoint detector subscription

Jim8y:
## Summary
- fix the Quick Start panel checkpoint detector subscription to use DetectorBase.onChange
- restore the VS Code extension TypeScript compile path

## Verification
- npm run compile
- npm run test:quickstart

## Notes
This is intentionally scoped to the compile-blocking event API mismatch.

### DIFF
--- a/.github/workflows/pr_extension.yml
+++ b/.github/workflows/pr_extension.yml
@@ -1,0 +1,42 @@
+name: On Pull Request (VS Code Extension)
+
+on:
+  pull_request:
+    branches:
+      - master
+      - development
+      - release/*
+    paths:
+      - '.github/workflows/pr_extension.yml'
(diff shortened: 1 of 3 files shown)


shargon:
What's the previous and new behavior?

ajara87 (approved):

shargon:
> What's the previous and new behavior?

@Jim8y please answer the comments before merge

Jim8y:
You're right, I should have answered before merging. Previous behavior: the Quick Start panel subscribed through the wrong detector event API, so checkpoint changes were not reliably picked up and the extension compile path was broken. New behavior: it subscribes through `DetectorBase.onChange`, so checkpoint changes trigger the panel update through the supported event surface.

PR #543, opened 2026-04-25
Keep WorkNet RPC bound to loopback by default

Jim8y:
## Summary
- change WorkNet RPC bind address from all interfaces to loopback
- keep the existing port and session configuration unchanged

## Verification
- dotnet build src/worknet/neoworknet.csproj --configuration Release

## Notes
The build succeeds with existing nullable warnings in bctklib/worknet; those are addressed separately in the warning-cleanup PR.


shargon (approved):
It's more secure

ajara87:
lgtm

PR #545, opened 2026-04-25
Fix notification pagination off-by-one

Jim8y:
## Summary
- make ExpressEnumNotifications return at most the requested take count
- preserve truncated=true detection when another notification exists

## Verification
- dotnet build src/neoxp/neoxp.csproj --configuration Release

## Notes
The build succeeds with existing nullable warnings in neoxp; this PR only fixes the pagination count behavior. Negative skip/take validation is split into a follow-up PR.

### DIFF
--- a/src/neoxp/Node/Modules/ExpressRpcServerPlugin.cs
+++ b/src/neoxp/Node/Modules/ExpressRpcServerPlugin.cs
@@ -318,7 +318,7 @@ namespace NeoExpress.Node
             var truncated = false;
             foreach (var (blockIndex, _, notification) in notifications)
             {
-                if (count++ > take)
+                if (count++ >= take)
                 {
                     truncated = true;
                     break;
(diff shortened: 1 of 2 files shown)


ajara87 (approved):

shargon (commented):
Lgtm

src/neoxp/Node/Modules/ExpressRpcServerPlugin.cs
```diff
@@ -343,6 +349,18 @@ public JObject ExpressEnumNotifications(JArray @params)
             };
         }
 
+        internal static (int skip, int take) GetNotificationPaging(JArray @params)
```
shargon, line 352: This method is only used once, I don't see any improvement with the change

Jim8y:
I extracted the helper mostly to make the pagination boundary testable without constructing the full RPC plugin path. The method is only used once, so I agree the production-code improvement is small; the value here was focused coverage for the off-by-one behavior.

PR #549, opened 2026-04-27
Validate stop node index before stopping

Jim8y:
## Summary
Fixes the CLI-1 fuzz finding by rejecting out-of-range `stop --node-index` values before indexing the consensus node list.

## Verification
- Direct CLI repro with `stop --node-index -1` returns `node-index must be in [0, 0]` instead of `ArgumentOutOfRangeException`.


ajara87 (approved):

shargon (commented):
Lgtm

PR #552, opened 2026-04-27
Treat null config settings as empty

Jim8y:
## Summary
Fixes the CLI-8 fuzz finding by preserving `settings: null` as an empty settings dictionary instead of allowing a later null-reference path.

## Verification
- `dotnet test test/test.bctklib/test.bctklib.csproj --filter ExpressChainTest`


shargon (approved):

ajara87:
lgtm

PR #554, opened 2026-04-27
Report malformed contract manifests cleanly

Jim8y:
## Summary
Fixes the MF-1/MF-2/MF-3 manifest parsing findings by wrapping malformed contract manifest parse failures with the manifest path and a stable input-error message.

## Verification
- Direct `contract hash` repro with `abi: null` no longer emits `NullReferenceException`, `InvalidCastException`, or `FormatException`.


shargon (approved):

ajara87:
lgtm

PR #556, opened 2026-04-27
Report malformed oracle response JSON cleanly

Jim8y:
## Summary
Fixes the ORC-1 fuzz finding by converting malformed `oracle response` JSON into a stable input error instead of leaking `JsonReaderException`.

## Verification
- Direct `oracle response` repro with `NaN` no longer emits `JsonReaderException`.


ajara87 (approved):

shargon (commented):
Same as others... better together

Jim8y:
Agreed. This should have been grouped with the other malformed-input error wrapping changes. I’ll avoid sending several tiny PRs for the same pattern.

PR #557, opened 2026-04-27
Report corrupt checkpoint archives cleanly

Jim8y:
## Summary
Fixes the CP-1 fuzz finding by wrapping corrupt checkpoint archive restore failures in a stable checkpoint input error.

## Verification
- Direct `checkpoint restore` repro with an empty archive no longer emits `InvalidDataException` or `EndOfStreamException`.


ajara87 (approved):

shargon (commented):
I think should be done in RockDbUtility

Jim8y:
Agreed. The restore boundary is really `RocksDbUtility`, so wrapping corrupt archive errors there would be cleaner than catching them at the caller. I missed that before merging; I’ll move that handling in a follow-up.

PR #558, opened 2026-04-27
Report overlong checkpoint paths cleanly

Jim8y:
## Summary
Fixes the CLI-3 long checkpoint name finding by converting `PathTooLongException` during checkpoint creation into a stable checkpoint path error. Existing relative subdirectory checkpoint paths remain supported.

## Verification
- Direct `checkpoint create` repro with a 400-character name no longer emits `PathTooLongException`.


ajara87 (approved):

shargon (commented):
Lgtm

PR #559, opened 2026-04-27
Let discard runs start without existing node state

Jim8y:
## Summary
Fixes the CLI-6 fuzz finding by letting `OpenForDiscard` handle an absent node data directory instead of pre-creating an empty RocksDB directory that leaks a missing `CURRENT` marker error.

## Verification
- Direct `run --discard` on a freshly created chain reaches `Neo express is running (CheckpointExpressStorage)` without RocksDB `CURRENT` errors.


shargon (approved):

ajara87:
lgtm

PR #561, opened 2026-04-28
Report missing storage contracts before update

Jim8y:
## Fuzzer context
This PR fixes BLOB-1, an issue identified by the neo-express fuzzer: `contract storage update` against a non-existent contract leaked `NullReferenceException`.

## Summary
- Validate that a target contract exists before applying `contract storage update`.
- Return `Contract not found` instead of leaking `NullReferenceException` for missing hashes.
- Covers BLOB-1.

## Verification
- Direct repro: `contract storage update 0x0101010101010101010101010101010101010101 key value --input default.neo-express` now returns `Contract not found` instead of `NullReferenceException`.

### DIFF
--- a/src/neoxp/Node/OfflineNode.cs
+++ b/src/neoxp/Node/OfflineNode.cs
@@ -411,10 +411,15 @@ namespace NeoExpress.Node
         public Task<int> PersistStorageKeyValueAsync(UInt160 scripthash, (string key, string value) storagePair)
             => MakeAsync(() =>
             {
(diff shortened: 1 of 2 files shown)


ajara87 (approved):

shargon (commented):
It's better to add the script hash in the exception message, otherwise it doesn't provide useful information

Jim8y:
Agreed. `Contract not found` is stable, but not useful enough when the caller passed a hash. I should have included the script hash in the message. I’ll fix that in a follow-up.

PR #562, opened 2026-04-28
Bound invocation file reads

Jim8y:
## Fuzzer context
This PR fixes HANG-1, an issue identified by the neo-express fuzzer: `contract invoke /dev/zero` could hang forever while reading an infinite stream as an invocation JSON file.

## Summary
- Add a 4 MiB read budget for `.neo-invoke.json` loading.
- Reject directories and oversized invocation files before parsing.
- Convert read-budget and JSON read failures into invocation-file errors.
- Covers HANG-1.

## Verification
- `dotnet test test/test.bctklib/test.bctklib.csproj --filter ContractParameterParserTest`
- Direct repro: `contract invoke /dev/zero genesis --input default.neo-express --results` now exits with `Invocation file /dev/zero is invalid: file is larger than 4194304 bytes` instead of timing out.

### DIFF
--- a/src/bctklib/ContractParameterParser.cs
+++ b/src/bctklib/ContractParameterParser.cs
@@ -25,6 +25,8 @@ namespace Neo.BlockchainToolkit
 {
     public class ContractParameterParser
     {
+        internal const long MaxInvocationFileBytes = 4 * 1024 * 1024;
+
         public delegate bool TryGetUInt160(string value, [MaybeNullWhen(false)] out UInt160 account);
 
         private readonly byte addressVersion;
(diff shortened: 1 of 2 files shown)


ajara87 (approved):

src/bctklib/ContractParameterParser.cs
```diff
@@ -50,22 +52,102 @@ public ContractParameterParser(byte addressVersion, TryGetUInt160? tryGetAccount
...
-
-            using var streamReader = fileSystem.File.OpenText(invokeFile);
+            if ((fileInfo.Attributes & FileAttributes.Directory) == FileAttributes.Directory)
+                throw new ArgumentException($"{path} is not a file", nameof(path));
+            if (fileInfo.Length > MaxInvocationFileBytes)
+                throw new Exception($"Invocation file {path} is invalid: file is larger than {MaxInvocationFileBytes} bytes");
+
+            using var stream = fileSystem.File.OpenRead(invokeFile);
+            using var budgetedStream = new ReadBudgetStream(stream, MaxInvocationFileBytes);
+            using var streamReader = new StreamReader(budgetedStream);
```
shargon, line 65: If was already checked the length why use a new class for it?

Jim8y:
`FileInfo.Length` catches regular oversized files, but it does not protect special files/streams such as `/dev/zero`, which can report length 0 and never reach EOF. The budgeted stream was added for that read-time guard. I agree it is heavier than a simple length check, but it covers the actual hang case.

PR #564, opened 2026-04-29
Treat unreadable execute input as invalid script

Jim8y:
## Summary

Fixes fuzzer-identified issue `EXEC-1`.

`neoxp execute <long invalid hex> --results` fell through script/base64 parsing and treated the raw argument as a file path. Very long fuzzed input could therefore leak `PathTooLongException` from the filesystem probe instead of reporting an invalid script.

This keeps existing script-file and NEF-file behavior, but treats unreadable or invalid file probes as a normal parse miss so the CLI returns `Invalid script: ...`.

## Validation

- `dotnet test test/test.workflowvalidation/test.workflowvalidation.csproj --filter ExecuteCommandTests`
- Direct fuzzer repro: long `0xb0...` execute input exits 1 with `Invalid script`, no `PathTooLongException`

### DIFF
--- a/src/neoxp/Commands/ExecuteCommand.cs
+++ b/src/neoxp/Commands/ExecuteCommand.cs
@@ -110,22 +110,32 @@ namespace NeoExpress.Commands
 
         private static Script? LoadFileScript(string fileName)
         {
-            var fileText = File.ReadAllText(fileName);
-            var txScript = ConvertTextToScript(fileText);
-            if (txScript != null)
-            {
-                return txScript;
-
-            }
-            var file = File.ReadAllBytes(fileName);
             try
             {
-                var nef = file.AsSerializable<NefFile>();
-                return nef?.Script;
+                if (!File.Exists(fileName))
+                    return null;
+
+                var fileText = File.ReadAllText(fileName);
+                var txScript = ConvertTextToScript(fileText);
+                if (txScript != null)
+                {
+                    return txScript;
+
+                }
+                var file = File.ReadAllBytes(fileName);
+                try
+                {
+                    var nef = file.AsSerializable<NefFile>();
+                    return nef?.Script;
+                }
+                catch (Exception)
+                {
+                }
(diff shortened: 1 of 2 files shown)


src/neoxp/Commands/ExecuteCommand.cs
```diff
@@ -110,22 +110,32 @@ internal async Task<int> OnExecuteAsync(CommandLineApplication app, IConsole con
...
+                    return txScript;
+
+                }
+                var file = File.ReadAllBytes(fileName);
+                try
+                {
+                    var nef = file.AsSerializable<NefFile>();
+                    return nef?.Script;
+                }
+                catch (Exception)
```
ajara87:
```suggestion
                catch (Exception ex) when (ex is FormatException or InvalidOperationException)
```

src/neoxp/Commands/ExecuteCommand.cs
```diff
@@ -110,22 +110,32 @@ internal async Task<int> OnExecuteAsync(CommandLineApplication app, IConsole con
...
-                return nef?.Script;
+                if (!File.Exists(fileName))
+                    return null;
+
+                var fileText = File.ReadAllText(fileName);
+                var txScript = ConvertTextToScript(fileText);
+                if (txScript != null)
+                {
+                    return txScript;
+
```
shargon:
```suggestion
```

src/neoxp/Commands/ExecuteCommand.cs
shargon: This try should be removed, is already catched

Jim8y:
Addressed the review feedback by removing the inner catch around NEF parsing. The outer invalid-input path now handles the parse exceptions and still reports the input as an invalid script. Local targeted validation passed (`ExecuteCommandTests`), and the PR checks are green.

Jim8y:
Combined into #580.

PR #565, opened 2026-04-29
Bound create batch file reads

Jim8y:
## Summary

Fixes fuzzer-identified issue `HANG-3`.

`neoxp create --batch-filename /dev/zero` created the chain and then attempted to load the batch file with `ReadAllLinesAsync`. Character devices and other infinite streams never reach EOF, so the command could hang in an unbounded read.

This adds a small create-time batch loader with a 4 MiB read budget. Normal batch files still produce the same string-array input for `BatchCommand`, while directories, oversized files, and streams exceeding the budget fail with a bounded validation error.

## Validation

- `dotnet test test/test.workflowvalidation/test.workflowvalidation.csproj --filter CreateCommandTests`
- Direct fuzzer repro: `timeout 10s neoxp create --output default.neo-express --batch-filename /dev/zero` exits 1 with bounded file-size error, not timeout 124

### DIFF
--- a/src/neoxp/Commands/CreateCommand.cs
+++ b/src/neoxp/Commands/CreateCommand.cs
@@ -48,6 +48,8 @@ namespace NeoExpress.Commands
         [Option(Description = "Use a batch file to initialize the blockchain after creation.")]
         internal string BatchFilename { get; set; } = string.Empty;
 
+        internal const long MaxBatchFileBytes = 4 * 1024 * 1024;
+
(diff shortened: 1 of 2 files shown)


shargon (approved):
Looks good, but to me it has no severity, it's a local attack, if someone try to load a big file by himself.. it's expected

Jim8y:
Fair point. I agree this is local robustness rather than a high-severity security issue. I treated it as a fuzzer hang guard because special files can block the command, but I should have described the severity more accurately.

PR #567, opened 2026-04-29
Bound oracle response file reads

Jim8y:
## Summary

Fixes fuzzer-identified issue `HANG-5`.

`neoxp oracle response https://x /dev/zero` parsed JSON directly from the supplied stream. Infinite streams never reach EOF, so the command could hang before JSON parsing returned.

This routes oracle response loading through a bounded helper: known oversized files are rejected up front, and streams are stopped after a 4 MiB budget. Existing malformed-JSON wrapping remains intact.

## Validation

- `dotnet test test/test.workflowvalidation/test.workflowvalidation.csproj --filter TransactionExecutorOracleResponseTests`
- Direct fuzzer repro: `timeout 10s neoxp oracle response https://x /dev/zero --input default.neo-express` exits 1 with bounded file-size error, not timeout 124
- Malformed JSON still exits 1 with `invalid JSON` error

### DIFF
--- a/src/neoxp/TransactionExecutor.cs
+++ b/src/neoxp/TransactionExecutor.cs
@@ -34,6 +34,8 @@ namespace NeoExpress
 
     class TransactionExecutor : IDisposable
     {
+        internal const long MaxOracleResponseFileBytes = 4 * 1024 * 1024;
+
         readonly ExpressChainManager chainManager;
         readonly IExpressNode expressNode;
         readonly IFileSystem fileSystem;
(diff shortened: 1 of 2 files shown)


shargon (commented):
Why neo express don't use the neo oracle?

Jim8y:
This PR is only about validating the local oracle response file before neo-express reads it into memory. It does not change the oracle protocol path. The file is user-provided input to the command, so neo-express still needs a local read limit before handing the bytes to the existing response handling flow.

Jim8y:
Combined into #580.

PR #572, opened 2026-04-29
Reject unsupported extension command URI arguments

Jim8y:
## Summary
- Treats this as a valid extension fuzzer finding: command URI arguments could include `path` or `blockchainIdentifier`, but the sanitizer silently dropped those unsupported fields.
- Rejects unsupported fields explicitly so malformed or hostile command invocations fail at the boundary instead of being partially interpreted.
- Adds focused unit coverage for supported inputs and the rejected unsupported fields.

## Fuzzer issue
- Fixes EXT-8 from the extension fuzzer findings.

## Validation
- `node --test -r ts-node/register src/extension/commands/commandArguments.test.ts`
- `npm run compile`
- `npm run test:quickstart`

### DIFF
--- a/extentions/neo3-visual-tracker/src/extension/commands/commandArguments.test.ts
+++ b/extentions/neo3-visual-tracker/src/extension/commands/commandArguments.test.ts
@@ -1,0 +1,18 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { sanitizeCommandArguments } from "./commandArguments";
+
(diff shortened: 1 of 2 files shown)


ajara87 (approved):

shargon (commented):
Why not group all this message changes or exception changes in only one Pr?

Jim8y:
Agreed. I split the extension command validation by fuzzer finding, but the review cost is higher than the benefit for small argument/message changes. I’ll group related exception/message changes into one PR going forward.

PR #574, opened 2026-04-29
Ignore custom RPC lists in untrusted workspaces

Jim8y:
## Summary
- Treats the workspace-controlled RPC list finding as valid for untrusted workspaces.
- Keeps built-in seed RPC URLs available, but ignores `neo-servers.json` files until VS Code reports the workspace as trusted.
- Leaves trusted-workspace behavior unchanged.

## Fuzzer issue
- Partially fixes EXT-6 from the extension fuzzer findings by adding a VS Code workspace-trust boundary around custom RPC URL discovery.

## Validation
- Red test first: `node --test -r ts-node/register src/extension/fileDetectors/serverListWorkspaceTrust.test.ts` failed because the trust helper did not exist.
- `node --test -r ts-node/register src/extension/fileDetectors/serverListWorkspaceTrust.test.ts`
- `npm run test:quickstart`
- `npm run compile`

## Residual risk
- This does not add TLS pinning or change behavior for already trusted workspaces. That is a broader product/security policy decision and should not be hidden inside this small PR.

### DIFF
--- a/extentions/neo3-visual-tracker/src/extension/fileDetectors/serverListDetector.ts
+++ b/extentions/neo3-visual-tracker/src/extension/fileDetectors/serverListDetector.ts
@@ -8,6 +8,7 @@ import IoHelpers from "../util/ioHelpers";
 import JSONC from "../util/JSONC";
 import Log from "../util/log";
 import posixPath from "../util/posixPath";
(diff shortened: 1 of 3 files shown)


ajara87 (approved):

shargon (commented):
Lgtm

PR #575, opened 2026-04-29
Report unbalanced quotes in batch files

Jim8y:
## Summary
- Treats the batch parser fuzzer finding as a valid low-severity UX issue.
- Reports an explicit parse error when a batch command line ends inside a quoted token.
- Adds focused parser tests for the malformed and still-valid quoted-token cases.

## Fuzzer issue
- Partially fixes BATCH from the fuzzer findings: unbalanced quotes are no longer silently consumed.
- This intentionally does not redesign escaping or argument length policy in the same PR.

## Validation
- Red test first: `dotnet test test/test.workflowvalidation/test.workflowvalidation.csproj --filter BatchCommandParserTests --no-restore` initially failed because the parser had no test-visible seam.
- `dotnet test test/test.workflowvalidation/test.workflowvalidation.csproj --filter BatchCommandParserTests --no-restore`
- `dotnet build src/neoxp/neoxp.csproj --no-restore`

### DIFF
--- a/src/neoxp/Commands/BatchCommand.cs
+++ b/src/neoxp/Commands/BatchCommand.cs
@@ -298,7 +298,7 @@ namespace NeoExpress.Commands
         }
 
         // SplitCommandLine method adapted from CommandLineStringSplitter class in https://github.com/dotnet/command-line-api
-        static IEnumerable<string> SplitCommandLine(string commandLine)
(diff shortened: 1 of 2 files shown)


ajara87 (approved):

shargon (commented):
Lgtm

PR #576, opened 2026-04-29
Document checkpoint zip-slip guard

Jim8y:
## Summary
- Documents the runtime path-containment guarantee used when restoring checkpoint archives.
- Keeps the existing `ZipFile.ExtractToDirectory` helper visible as the security boundary for zip-slip attempts.

## Fuzzer issue
- Addresses CP-2, which was confirmed as a security non-finding: .NET rejects archive entries that would escape the destination directory.
- This is intentionally documentation-only so future refactors do not accidentally replace the safe helper with manual extraction.

## Validation
- `dotnet build src/bctklib/bctklib.csproj --no-restore`

### DIFF
--- a/src/bctklib/persistence/RocksDbUtility.cs
+++ b/src/bctklib/persistence/RocksDbUtility.cs
@@ -203,6 +203,8 @@ namespace Neo.BlockchainToolkit.Persistence
 
             static void ExtractCheckpoint(string checkPointArchive, string restorePath)
             {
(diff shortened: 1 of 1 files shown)


ajara87 (approved):

shargon (commented):
Only two lines?

Jim8y:
Yes, intentionally small. The finding was a non-issue because `ZipFile.ExtractToDirectory` already enforces the path boundary, so I only added a note to keep that security boundary visible. I should have explained that in the PR before merging.

PR #580, opened 2026-04-30
Harden file input validation

Jim8y:
## Summary
- Treat unreadable or oversized execute input files as invalid scripts instead of loading them or throwing raw file errors.
- Bound oracle response file reads before parsing local response JSON.
- Add targeted workflow validation coverage for both cases.

## Validation
- dotnet test test/test.workflowvalidation/test.workflowvalidation.csproj --filter "FullyQualifiedName~ExecuteCommandTests|FullyQualifiedName~TransactionExecutorOracleResponseTests"

Supersedes #564 and #567.

### DIFF
--- a/src/neoxp/Commands/ExecuteCommand.cs
+++ b/src/neoxp/Commands/ExecuteCommand.cs
@@ -26,6 +26,7 @@ namespace NeoExpress.Commands
     [Command(Name = "execute", Description = "Invoke a custom script, the input text will be converted to script with a priority: hex, base64, file path.")]
     class ExecuteCommand
     {
+        internal const long MaxExecuteScriptFileBytes = 4 * 1024 * 1024;
 
         readonly ExpressChainManagerFactory chainManagerFactory;
         readonly TransactionExecutorFactory txExecutorFactory;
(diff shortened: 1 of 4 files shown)


src/neoxp/Commands/ExecuteCommand.cs
```diff
@@ -110,22 +110,25 @@ internal async Task<int> OnExecuteAsync(CommandLineApplication app, IConsole con
...
-            var txScript = ConvertTextToScript(fileText);
-            if (txScript != null)
-            {
-                return txScript;
-
-            }
-            var file = File.ReadAllBytes(fileName);
             try
             {
+                if (!File.Exists(fileName))
```
Wi1l-B0t, line 116: Check file size?

Do not load large file?

Jim8y: Added a `MaxExecuteScriptFileBytes` guard. The command now checks `FileInfo.Length` before `ReadAllText` / `ReadAllBytes`, so oversized files are treated as invalid scripts without loading them into memory. Added a targeted workflow validation test for the oversized-file path.

src/neoxp/TransactionExecutor.cs
```diff
@@ -464,6 +450,61 @@ public async Task OracleResponseAsync(string url, string responsePath, ulong? re
...
+            try
+            {
+                using var stream = fileSystem.File.OpenRead(responsePath);
+                using var memory = new MemoryStream();
+                var buffer = new byte[8192];
+                long bytesRead = 0;
+
+                while (true)
+                {
+                    var read = await stream.ReadAsync(buffer, 0, buffer.Length, token).ConfigureAwait(false);
```
Wi1l-B0t: Use `CopyToAsync` instead of coping manually?

Jim8y: I kept the manual loop so the read path can enforce `MaxOracleResponseFileBytes` while copying. `CopyToAsync` would rely only on the `FileInfo.Length` pre-check and would not catch a file that changes after the check.

Wi1l-B0t (dismissed):

shargon (dismissed):

src/neoxp/TransactionExecutor.cs
Wi1l-B0t: > I kept the manual loop so the read path can enforce `MaxOracleResponseFileBytes` while copying. `CopyToAsync` would rely only on the `FileInfo.Length` pre-check and would not catch a file that changes after the check.

I don't think there's any need to worry about file changes after the lengthcheck.
 If we have to consider this, then many other operations would need to be redone, not just here.

ajara87 (dismissed):

src/neoxp/TransactionExecutor.cs
shargon: I said this in other pr, I think the same as @Wi1l-B0t

Wi1l-B0t (approved):

ajara87 (approved):

shargon (approved):

PR #589, opened 2026-05-15
Clear persistent tracking changes on reset

Jim8y:
## Summary
- Delete all persisted tracking entries when PersistentTrackingStore.Reset is called
- Add coverage that reset restores underlying values, removes added entries, and clears deletions

## Testing
- dotnet test test/test.bctklib/test.bctklib.csproj --filter persistent_tracking_reset_discards_tracked_changes
- dotnet test test/test.bctklib/test.bctklib.csproj
- dotnet format neo-express.sln --verify-no-changes --no-restore --verbosity minimal
- dotnet build neo-express.sln --configuration Release

### DIFF
--- a/src/bctklib/persistence/PersistentTrackingStore.cs
+++ b/src/bctklib/persistence/PersistentTrackingStore.cs
@@ -23,7 +23,8 @@ namespace Neo.BlockchainToolkit.Persistence
         readonly static ReadOnlyMemory<byte> DELETED_PREFIX = (new byte[] { DELETED_KEY }).AsMemory();
 
         readonly RocksDb db;
-        readonly ColumnFamilyHandle columnFamily;
+        ColumnFamilyHandle columnFamily;
+        readonly string? columnFamilyName;
         readonly IReadOnlyStore<byte[], byte[]> store;
         readonly bool shared;
         bool disposed;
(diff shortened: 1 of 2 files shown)


ajara87 (dismissed):

src/bctklib/persistence/PersistentTrackingStore.cs
```diff
@@ -61,7 +61,14 @@ public void Reset()
         {
             if (disposed || db.Handle == IntPtr.Zero)
                 throw new ObjectDisposedException(nameof(RocksDbStore));
+
             using var iterator = db.NewIterator(columnFamily);
+            using var batch = new WriteBatch();
```
shargon, line 84: Could you explain me the expected behavior, it's a reset for all the store, or for what? It's used? A summary should be good

src/bctklib/persistence/PersistentTrackingStore.cs
```diff
@@ -61,7 +61,14 @@ public void Reset()
...
+            for (iterator.SeekToFirst(); iterator.Valid(); iterator.Next())
```
Wi1l-B0t, line 85: It's inefficient.

The `leveldb` has a `DestroyDB` API.

src/bctklib/persistence/PersistentTrackingStore.cs
Jim8y: Reset only clears the PersistentTrackingStore overlay so reads fall back to the backing store state. It is used by worknet reset to discard persisted local tracking changes without deleting the underlying store data.

src/bctklib/persistence/PersistentTrackingStore.cs
Jim8y: DestroyDB would remove the whole RocksDB database. PersistentTrackingStore can share the database with other data, so reset now drops only its owned column family when available and otherwise deletes tracked keys from that tracking column family.

Wi1l-B0t (approved):

ajara87 (approved):

shargon (approved):

PR #594, opened 2026-05-15
Fix iterator JSON output in test runner

Jim8y:
## Summary
- Close iterator JSON values with the matching object terminator
- Add test coverage for serializing iterator stack items in Neo.Test.Runner

## Testing
- dotnet test test/test-runner/test-runner.csproj --configuration Release --no-restore --verbosity minimal
- dotnet test neo-express.sln --configuration Release --no-restore --verbosity minimal --filter "FullyQualifiedName!~test.workflowvalidation"
- git diff --check


test/test-runner/test-runner.csproj
```diff
@@ -0,0 +1,34 @@
...
+    </PackageReference>
+    <PackageReference Include="xunit.v3" Version="3.1.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\runner\runner.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Update="Nerdbank.GitVersioning" Version="3.8.118" />
```
shargon: Is required?

shargon (commented):
Good catch

Wi1l-B0t (dismissed):

test/test-runner/test-runner.csproj
ajara87: not needed

shargon (approved):

ajara87 (approved):

PR #599, opened 2026-05-16
Emit JSON for candidate list

Jim8y:
## Summary
- Return structured candidate records from `ListCandidatesAsync` instead of preformatted text.
- Make `neoxp candidate list --json` emit a JSON array with public keys and votes.
- Preserve the existing text output layout when `--json` is not set.

## Validation
- `dotnet test test/test.workflowvalidation/test.workflowvalidation.csproj --configuration Release --no-build --no-restore --verbosity minimal --filter CandidateListCommandTests`
- `dotnet test neo-express.sln --configuration Release --no-restore --verbosity minimal --filter "FullyQualifiedName!~test.workflowvalidation.WorkflowValidationTests&FullyQualifiedName!~test.workflowvalidation.NeoxpToolIntegrationTests&FullyQualifiedName!~test.workflowvalidation.NeoxpAdvancedIntegrationTests"`
- `git diff --check`

### DIFF
--- a/src/neoxp/Commands/CandidateCommand.List.cs
+++ b/src/neoxp/Commands/CandidateCommand.List.cs
@@ -9,6 +9,8 @@
 // modifications are permitted.
 
 using McMaster.Extensions.CommandLineUtils;
+using Newtonsoft.Json;
+using System.Globalization;
 
 namespace NeoExpress.Commands
 {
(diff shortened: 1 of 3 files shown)


src/neoxp/Commands/CandidateCommand.List.cs
```diff
@@ -9,6 +9,8 @@
 // modifications are permitted.
 
 using McMaster.Extensions.CommandLineUtils;
+using Newtonsoft.Json;
```
shargon: Can we use system.json?

ajara87: agree

shargon (approved):

ajara87 (approved):

PR #601, opened 2026-05-16
Write transaction hashes as JSON strings

Jim8y:
## Summary
- Make JSON transaction hash output valid JSON by writing it through `JsonTextWriter`.
- Keep the non-JSON transaction message unchanged.
- Add a regression test that parses the JSON transaction hash output.

## Validation
- `dotnet test test/test.workflowvalidation/test.workflowvalidation.csproj --configuration Release --no-restore --verbosity minimal --filter TextWriterExtensionsTests`
- `dotnet test neo-express.sln --configuration Release --no-restore --verbosity minimal --filter "FullyQualifiedName!~test.workflowvalidation.WorkflowValidationTests&FullyQualifiedName!~test.workflowvalidation.NeoxpToolIntegrationTests&FullyQualifiedName!~test.workflowvalidation.NeoxpAdvancedIntegrationTests"`
- `git diff --check`

### DIFF
--- a/src/neoxp/Extensions/Extensions.cs
+++ b/src/neoxp/Extensions/Extensions.cs
@@ -239,8 +239,7 @@ namespace NeoExpress
         {
             if (json)
             {
-                using var jsonWriter = new Newtonsoft.Json.JsonTextWriter(writer);
-                await jsonWriter.WriteValueAsync(txHash.ToString()).ConfigureAwait(false);
+                await writer.WriteAsync($"\"{txHash}\"").ConfigureAwait(false);
             }
             else
             {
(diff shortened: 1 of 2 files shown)


src/neoxp/Extensions/Extensions.cs
```diff
@@ -239,7 +239,8 @@ public static async Task WriteTxHashAsync(this TextWriter writer, UInt256 txHash
         {
             if (json)
             {
-                await writer.WriteLineAsync($"{txHash}").ConfigureAwait(false);
+                using var jsonWriter = new Newtonsoft.Json.JsonTextWriter(writer);
+                await jsonWriter.WriteValueAsync(txHash.ToString()).ConfigureAwait(false);
```
shargon: Maybe we can just add the quotes, it will be faster

Wi1l-B0t (approved):

shargon (approved):

PR #603, opened 2026-05-16
Use default C# language version

Jim8y:
## Summary
- Remove the explicit C# 10 language version override from src/Directory.Build.props.
- Let the net10.0 target framework and SDK select the default C# language version.
- Include props and targets files in the C# PR workflow path filters so build configuration changes are validated.

## Validation
- dotnet build src/neoxp/neoxp.csproj --no-restore --verbosity minimal
- dotnet msbuild src/neoxp/neoxp.csproj -getProperty:LangVersion -> 14.0
- dotnet msbuild src/bctklib/bctklib.csproj -getProperty:LangVersion -> 14.0

### DIFF
--- a/.github/workflows/pr_cs.yml
+++ b/.github/workflows/pr_cs.yml
@@ -11,6 +11,8 @@ on:
       - '.github/workflows/test.yml'
       - '**.cs'
       - '**.csproj'
+      - '**.props'
+      - '**.targets'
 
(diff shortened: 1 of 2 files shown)


Wi1l-B0t (approved):

shargon (approved):
I prefer 'latest', but default it's better than 10

ajara87 (approved):

PR #612, opened 2026-05-30
Release resources in RestoreCheckpoint

ajara87:
# Summary
Temporary resource not released in `RestoreCheckpoint` if a post-creation exception occurs.

# Problem
The flow does not adequately protect the state if `fileSystem.Directory.Delete(nodePath, true)` or `fileSystem.Directory.Move(checkpointTempPath, nodePath)` fail:

- If `Delete(nodePath, true)` throws an exception after partially deleting files, folderCleanup is caught but nodePath is left in an inconsistent state.

- If `Move(checkpointTempPath, nodePath)` fails (example,  due to permissions) the `checkpointTempPath` folder is properly cleaned up by the disposable but nodePath may have been deleted without being able to restore the checkpoint leaving the blockchain in a corrupted state.

# Tests
- RestoreCheckpoint_RecoveryOnMoveFail_RestoresBackup
- RestoreCheckpoint_CleanupBackupOnSuccess

### DIFF
--- a/src/neoxp/ExpressChainManager.cs
+++ b/src/neoxp/ExpressChainManager.cs
@@ -248,7 +248,7 @@ namespace NeoExpress
 
             var checkpointTempPath = fileSystem.GetTempFolder();
             var nodePath = fileSystem.GetNodePath(node);
-            var nodePathBackup = string.Concat(nodePath, ".backup-", Guid.NewGuid().ToString().AsSpan(0, 8));
+            var nodePathBackup = string.Concat(nodePath, ".backup-", Guid.NewGuid().ToString("N"));
 
             // Step 1: Restore checkpoint to temp location
             var wallet = DevWallet.FromExpressWallet(ProtocolSettings, node.Wallet);
--- a/src/neoxp/ExpressChainManager.cs
+++ b/src/neoxp/ExpressChainManager.cs
@@ -248,24 +248,15 @@ namespace NeoExpress
 
             var checkpointTempPath = fileSystem.GetTempFolder();
             var nodePath = fileSystem.GetNodePath(node);
-            var nodePathBackup = nodePath + ".backup-" + Guid.NewGuid().ToString().Substring(0, 8);
+            var nodePathBackup = string.Concat(nodePath, ".backup-", Guid.NewGuid().ToString().AsSpan(0, 8));
+
+            // Step 1: Restore checkpoint to temp location
+            var wallet = DevWallet.FromExpressWallet(ProtocolSettings, node.Wallet);
+            var multiSigAccount = wallet.GetMultiSigAccounts().Single();
+            RocksDbUtility.RestoreCheckpoint(checkPointArchive, checkpointTempPath, ProtocolSettings.Network, ProtocolSettings.AddressVersion, multiSigAccount.ScriptHash);
 
             try
             {
-                // Step 1: Restore checkpoint to temp location
-                using var folderCleanup = AnonymousDisposable.Create(() =>
-                {
-                    if (fileSystem.Directory.Exists(checkpointTempPath))
-                    {
-                        fileSystem.Directory.Delete(checkpointTempPath, true);
-                    }
-                });
-
-                var wallet = DevWallet.FromExpressWallet(ProtocolSettings, node.Wallet);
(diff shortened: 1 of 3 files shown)


src/neoxp/ExpressChainManager.cs
```diff
@@ -247,30 +247,67 @@ public void RestoreCheckpoint(string checkPointArchive, bool force)
...
+
+            try
             {
-                if (fileSystem.Directory.Exists(checkpointTempPath))
+                // Step 1: Restore checkpoint to temp location
+                using var folderCleanup = AnonymousDisposable.Create(() =>
+                {
+                    if (fileSystem.Directory.Exists(checkpointTempPath))
+                    {
+                        fileSystem.Directory.Delete(checkpointTempPath, true);
```
Wi1l-B0t: Fsync is needed after each modification operation for data safety.

ajara87: The `IFileSystem` abstraction used here doesn't expose fsync and these are directory-level move/delete operations rather than file writes

shargon: This should remove `nodePathBackup`, otherwise it will remove other checkpoints

ajara87: > This should remove `nodePathBackup`, otherwise it will remove other checkpoints

replace AnonymousDisposable with an explicit finally block that cleans up only checkpointTempPath making the inner catch the sole recovery point for nodePathBackup

src/neoxp/ExpressChainManager.cs
```diff
@@ -247,30 +247,59 @@ public void RestoreCheckpoint(string checkPointArchive, bool force)
             }
 
             var checkpointTempPath = fileSystem.GetTempFolder();
-            using var folderCleanup = AnonymousDisposable.Create(() =>
+            var nodePath = fileSystem.GetNodePath(node);
+            var nodePathBackup = string.Concat(nodePath, ".backup-", Guid.NewGuid().ToString().AsSpan(0, 8));
```
cschuchardt88:
```suggestion
            var nodePathBackup = string.Concat(nodePath, ".backup-", Guid.NewGuid().ToString("N"));
```

Output: `.backup-1a2b3c4d5e6f7a8b9c0d1e2f3a4b5c6d`

cschuchardt88 (changes requested):

src/neoxp/ExpressChainManager.cs
Jim8y: Updated to use the full N-format GUID suffix.

Jim8y (approved):
LGTM

Wi1l-B0t (approved):

src/neoxp/ExpressChainManager.cs
```diff
@@ -247,30 +247,59 @@ public void RestoreCheckpoint(string checkPointArchive, bool force)
...
+                // Step 4: Clean up backup if successful
+                if (fileSystem.Directory.Exists(nodePathBackup))
+                {
+                    fileSystem.Directory.Delete(nodePathBackup, true);
+                }
+            }
+            finally
+            {
+                // Clean up temp folder if it still exists
+                if (fileSystem.Directory.Exists(checkpointTempPath))
```
shargon, line 298: Here is deleting all the temporary files in the system

shargon (changes requested):

src/neoxp/ExpressChainManager.cs
ajara87: `checkpointTempPath` is a unique path like %TEMP%/abc123. The finally deletes only that directory.

shargon:
```
var checkpointTempPath = fileSystem.GetTempFolder();
```
This is not the system temp folder?

ajara87: No, `GetTempFolder()` returns a randomly named directory not a file and not the system temp root. The finally block deletes only that specific directory

shargon (approved):

src/neoxp/ExpressChainManager.cs
shargon: You are right!

PR #621, opened 2026-06-21
Fix branch coverage counter key mismatch in TestApplicationEngine

Jim8y:
## Problem

`TestApplicationEngine.PostExecuteInstruction` records conditional-branch coverage into a per-contract `Dictionary<int, (int branchCount, int continueCount)>`, but it **reads** and **writes** that dictionary with different keys:

```csharp
var (hash, branchIP, offsetIP) = branchInstructionInfo;
...
// read uses offsetIP (the branch *target* address)
var (branchCount, continueCount) = branchMap.TryGetValue(offsetIP, out var value)
    ? value : (0, 0);

if (currentIP == offsetIP)
    branchMap[branchIP] = (branchCount + 1, continueCount);   // write uses branchIP
else if (currentIP == branchIP)
    branchMap[branchIP] = (branchCount, continueCount + 1);   // write uses branchIP
```

Because the read key (`offsetIP`) never matches the write key (`branchIP`), the read-modify-write never accumulates: every branch event reads a default `(0, 0)` from the unrelated `offsetIP` slot and overwrites the `branchIP` slot, so the stored tuple only ever reflects the **last** event rather than the running total. The coverage consumer (`ContractCoverageCollector.RecordBranch` / the raw coverage format) keys branch hits by the branch instruction address, matching the write side — so the emitted branch-coverage counts are wrong.

## Fix

Read the accumulator with `branchIP`, so the read and write operate on the same key and the counters accumulate correctly.

## Verification

- New regression test `BranchCoverageTests` executes a loop whose single conditional branch is taken twice and asserts the recorded branch-taken count is `2`. With the bug present the test fails (`found 0`); with the fix it passes.
- `dotnet test test/test.bctklib` — all 253 tests pass (ran 3× to confirm no flakiness).

The test passes a `MockFileSystem` to the engine so it is unaffected by the process-global coverage-path environment variable other engine tests set.

### DIFF
--- a/src/bctklib/smart-contract/TestApplicationEngine.cs
+++ b/src/bctklib/smart-contract/TestApplicationEngine.cs
@@ -273,7 +273,7 @@ namespace Neo.BlockchainToolkit.SmartContract
                 branchMaps.Add(branchInstructionInfo.ContractHash, branchMap);
             }
 
-            var (branchCount, continueCount) = branchMap.TryGetValue(offsetIP, out var value)
+            var (branchCount, continueCount) = branchMap.TryGetValue(branchIP, out var value)
                 ? value : (0, 0);
 
             if (currentIP == offsetIP)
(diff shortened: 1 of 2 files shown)


Wi1l-B0t (approved):
Minor Fix.

ajara87 (approved):

shargon (commented):
LGTM

PR #622, opened 2026-06-21
Throw a clear error when an oracle response references a missing transaction

Jim8y:
## Problem

`NodeUtility.CreateResponseTx` resolves the oracle request's original transaction and then uses its block index:

```csharp
var requestTx = NativeContract.Ledger.GetTransactionState(snapshot, request.OriginalTxid);
...
ValidUntilBlock = requestTx.BlockIndex + settings.MaxValidUntilBlockIncrement,
```

`LedgerContract.GetTransactionState` returns `null` when the transaction is not present in the ledger. If an oracle response references an original transaction that does not exist, `requestTx.BlockIndex` dereferences `null` and the caller sees an opaque `NullReferenceException` rather than a meaningful error. This path is reachable from both `OfflineNode` and the `ExpressRpcServerPlugin` oracle-response handling.

## Fix

Guard the lookup and throw a descriptive exception that names the missing transaction — consistent with the existing null-guarded lookup in `OfflineNode.GetTransactionHeight` (which already uses `GetTransactionState(...)?.BlockIndex`).

```csharp
if (requestTx is null)
    throw new Exception($"Could not find the original transaction {request.OriginalTxid} for the oracle response.");
```

## Verification

- New regression test `NodeUtilityOracleResponseTests` calls `CreateResponseTx` against an empty ledger and asserts the result is the descriptive error and **not** a `NullReferenceException`. With the guard removed the test fails (the call throws `NullReferenceException`); with the fix it passes.
- `dotnet build neo-express.sln -c Release` succeeds.
- CI-filtered `dotnet test` across the solution passes (test.workflowvalidation 61, all others green).

### DIFF
--- a/src/neoxp/Node/NodeUtility.cs
+++ b/src/neoxp/Node/NodeUtility.cs
@@ -142,6 +142,8 @@ namespace NeoExpress.Node
                 throw new Exception("No oracle nodes available. Have you enabled oracles via the `oracle enable` command?");
 
             var requestTx = NativeContract.Ledger.GetTransactionState(snapshot, request.OriginalTxid);
+            if (requestTx is null)
+                throw new Exception($"Could not find the original transaction {request.OriginalTxid} for the oracle response.");
             var n = oracleNodes.Count;
             var m = n - (n - 1) / 3;
             var oracleSignContract = Contract.CreateMultiSigContract(m, oracleNodes);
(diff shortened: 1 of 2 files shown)


Wi1l-B0t (approved):
Minor Fix.

ajara87 (approved):
lgtm

shargon (commented):
LGTM

PR #627, opened 2026-06-21
Dispose the cancellation token source in ExpressRpcServerPlugin

Jim8y:
## Problem

`ExpressRpcServerPlugin` owns a `CancellationTokenSource`:

```csharp
readonly CancellationTokenSource cancellationToken = new();
```

`ExpressShutdown` arms it with `cancellationToken.CancelAfter(...)` to schedule node shutdown, which starts an internal timer. But `Dispose` only disposed the RPC server:

```csharp
public override void Dispose()
{
    rpcServer?.Dispose();
    base.Dispose();
}
```

The token source was never disposed, leaking its timer registration.

## Fix

Dispose the token source in `Dispose`. The plugin is created with a `using` declaration in `ExpressChainManager`, and its token is consumed via a linked token source whose `WaitHandle.WaitOne()` has already returned by the time the `using` scope unwinds and `Dispose` runs — so disposal is safe at that point. `CancellationTokenSource.Dispose` is also idempotent.

## Verification

- `dotnet build src/neoxp -c Release` succeeds.

Note: the plugin requires a live `NeoSystem` to instantiate, so it is exercised by the integration suites rather than the fast unit tests; this is a one-line lifecycle fix to an existing `Dispose` override.

### DIFF
--- a/src/neoxp/Node/Modules/ExpressRpcServerPlugin.cs
+++ b/src/neoxp/Node/Modules/ExpressRpcServerPlugin.cs
@@ -69,6 +69,7 @@ namespace NeoExpress.Node
         public override void Dispose()
         {
             rpcServer?.Dispose();
+            cancellationToken.Dispose();
             base.Dispose();
         }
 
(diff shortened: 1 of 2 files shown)


Wi1l-B0t (commented):
Minor Fix.

shargon (approved):

ajara87 (approved):

PR #631, opened 2026-06-21
Honor the because reason in StorageItemAssertions.Be

Jim8y:
## Problem

`StorageItemAssertions.Be<T>` builds a failure message containing the `{reason}` placeholder, but never calls `BecauseOf`:

```csharp
Execute.Assertion
    .Given(() => convert(Subject.Value))
    .ForCondition(subject => subject.Equals(expected))
    .FailWith("Expected {context:StorageItem} to be {0}{reason}, but was {1}.",
        _ => expected, subject => subject);
```

Without `BecauseOf`, the `{reason}` placeholder is always empty, so any `because`/`becauseArgs` a caller passes to `Be(...)` are silently dropped from the failure message. The sibling `StackItemAssertions` applies `BecauseOf` to its assertions.

## Fix

Apply `.BecauseOf(because, becauseArgs)` to the assertion chain so the supplied reason is rendered into the message.

## Verification

- New test `Be_includes_the_supplied_reason_in_the_failure_message` asserts the custom reason appears in the failure message. With `BecauseOf` removed the test fails; with the fix it passes.
- `dotnet test test/test-assertions` — all 18 tests pass.

### DIFF
--- a/src/assertions/StorageItemAssertions.cs
+++ b/src/assertions/StorageItemAssertions.cs
@@ -28,6 +28,7 @@ namespace Neo.Assertions
             where T : IEquatable<T>
         {
             Execute.Assertion
+                .BecauseOf(because, becauseArgs)
                 .Given(() => convert(Subject.Value))
                 .ForCondition(subject => subject.Equals(expected))
                 .FailWith("Expected {context:StorageItem} to be {0}{reason}, but was {1}.",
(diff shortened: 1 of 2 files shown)


Wi1l-B0t (approved):
Minor Fix.

ajara87 (approved):
lgtm

shargon (commented):
LGTM

PR #637, opened 2026-06-22
Await the .NET check before launching neoxp

Jim8y:
## Problem

`checkForDotNet` is `async`, but both call sites test its result synchronously:

```ts
if (!this.checkForDotNet()) {
  return ...;
}
```

The method returns a `Promise`, which is always truthy, so `!this.checkForDotNet()` is always `false`. As a result the check never gates the launch: the `dotnet --version` probe runs but its result (and its 60-second cache) is discarded, and the "download .NET" message shown when .NET is missing or too old never fires. neoxp is launched regardless, so a missing/old .NET surfaces as a raw spawn failure instead of the intended dialog.

## Fix

`await` the call at both sites. Both are inside `async` methods (`runInTerminal` and `runUnsafe`), so the check now runs and gates the launch as intended.

## Verification

- `npx tsc --noEmit` passes (exit 0).
- `eslint` reports no new warnings on the changed file (the 4 remaining warnings are pre-existing, outside the changed lines).
- `npm run test:quickstart` passes (4/4).

### DIFF
--- a/extentions/neo3-visual-tracker/src/extension/neoExpress/neoExpress.ts
+++ b/extentions/neo3-visual-tracker/src/extension/neoExpress/neoExpress.ts
@@ -63,7 +63,7 @@ export default class NeoExpress {
   }
 
   async runInTerminal(name: string, command: Command, ...options: string[]) {
-    if (!(await this.checkForDotNet())) {
+    if (!(await this.checkForDotNetAsync())) {
       return null;
     }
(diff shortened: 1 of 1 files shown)


shargon (commented):
Could you rename the method to "Async"?

ajara87:
> Could you rename the method to "Async"?

agree with @shargon

ajara87 (approved):

shargon (approved):

PR #641, opened 2026-06-22
Decode binary contract parameters on the heap

Jim8y:
## Problem

`ContractParameterParser.ParseBinary` sizes a `stackalloc` buffer from the length of the input string:

```csharp
var value = json.Value<string>() ?? "";
Span<byte> span = stackalloc byte[value.Length / 4 * 3];
```

`value` comes from a parameter or invocation file (reachable via `LoadInvocationScriptAsync` / `ParseObjectParameter`). The whole-file 4 MiB cap does not bound an individual string value, so a value of a few hundred KB or more produces a multi-hundred-KB+ `stackalloc` and overflows the stack. `StackOverflowException` is uncatchable — it crashes the process — and the surrounding `try/catch` only filters `JsonException`/`FormatException`.

## Fix

Allocate the decode buffer on the heap (`new byte[...]`). Behavior is otherwise unchanged: base64 is attempted first, then hex.

## Verification

- New regression test `ParseObjectParameter_bytearray_large_base64` parses a 4 MiB byte array encoded as base64. With the previous `stackalloc` the test process crashes (stack overflow, exit code 134); with the heap buffer it passes.
- `dotnet test test/test.bctklib` — all 253 tests pass.

### DIFF
--- a/src/bctklib/ContractParameterParser.cs
+++ b/src/bctklib/ContractParameterParser.cs
@@ -519,10 +519,13 @@ namespace Neo.BlockchainToolkit
             static byte[] ParseBinary(JToken json)
             {
                 var value = json.Value<string>() ?? "";
-                Span<byte> span = stackalloc byte[value.Length / 4 * 3];
-                if (Convert.TryFromBase64String(value, span, out var written))
+                // Decode on the heap rather than via stackalloc: the buffer length is
+                // derived from the (untrusted) input string, so a large value would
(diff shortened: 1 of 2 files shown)


shargon (approved):

Wi1l-B0t (approved):
Minor Fix.

ajara87 (commented):
lgtm

PR #646, opened 2026-06-22
Guard state-service download loops against a non-advancing cursor

Jim8y:
## Problem

`StateServiceStore` pages through `FindStates` results, advancing the query cursor to the last key of each truncated page:

```csharp
from = found.Results[^1].key;
```

This happens in the shared `WriteFoundStates` helper (used by `DownloadStatesAsync`/`DownloadStates`) and in the `GetContractsAsync` loop. Neither checks that the cursor actually advanced. `FindStates` returns keys in ascending order, so a well-behaved node always moves the cursor forward — but a malfunctioning or malicious state node that returns a truncated page with the same (or an earlier) last key drives the loop forever, re-fetching the same page and accumulating state without bound (an unbounded-memory / hang DoS against any worknet/neotrace consumer pointed at it).

## Fix

Route both cursor updates through a single guard, `EnsureCursorAdvanced`, which rejects a cursor that does not move strictly forward (lexicographic byte order):

```csharp
if (((ReadOnlySpan<byte>)next).SequenceCompareTo(previous) <= 0)
    throw new InvalidOperationException("...did not advance the query cursor");
```

A legitimate node never trips this, since each truncated page's last key is strictly greater than the cursor it was queried with.

## Verification

- New tests `StateServiceStoreCursorTests` cover the guard: advancing cursors (including from an empty/shorter prefix) are returned; an identical, backward, or shorter-prefix cursor throws. Weakening the guard makes the rejection test fail, confirming it is meaningful.
- `dotnet test test/test.bctklib` — all 254 tests pass (the `out`→`ref` signature change and both call sites compile and regress nothing).

Note: the cursor logic sits behind MPT-proof validation and the buggy behavior is an infinite loop, so it cannot be exercised end-to-end in a focused unit test; the advance check is extracted into a pure, directly-tested helper and applied at both cursor sites.

### DIFF
--- a/src/bctklib/persistence/StateServiceStore.cs
+++ b/src/bctklib/persistence/StateServiceStore.cs
@@ -201,7 +201,7 @@ namespace Neo.BlockchainToolkit.Persistence
                     }
                     if (!found.Truncated || found.Results.Length == 0)
                         break;
-                    from = found.Results[^1].key;
+                    from = EnsureCursorAdvanced(from, found.Results[^1].key);
                 }
                 return contracts;
             }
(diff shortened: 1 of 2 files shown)


shargon (approved):

Wi1l-B0t (approved):

ajara87 (commented):
lgtm

PR #655, opened 2026-06-22
Format Cobertura coverage rates with the invariant culture

Jim8y:
## Problem

The Cobertura report formats `line-rate`, `branch-rate`, `condition-coverage`, and `coverage` values with interpolated strings and no format provider, so they use the **current culture**:

```csharp
writer.WriteAttributeString("line-rate", $"{lineRate:N4}");
```

On a machine with a non-invariant culture (e.g. de-DE, which uses `,` as the decimal separator) the report emits `line-rate="0,6667"`, which is invalid for the Cobertura XML schema and is misparsed by coverage tooling.

## Fix

Wrap the numeric formatting with `FormattableString.Invariant(...)` at every rate/percentage site in both Cobertura writers, so the values always use `.` regardless of host culture. (Adds the missing `using System;` to the partial writer.)

## Verification

- New test `CoberturaCultureTests` sets `CurrentCulture` to de-DE, generates the report, and asserts the root `line-rate`/`branch-rate` attributes contain no comma and use `.`. Reverting any rate to culture-sensitive formatting fails the test.
- `dotnet test test/test-collector` passes.

### DIFF
--- a/src/collector/Formats/CoberturaFormat.ContractCoverageWriter.cs
+++ b/src/collector/Formats/CoberturaFormat.ContractCoverageWriter.cs
@@ -9,6 +9,7 @@
 // modifications are permitted.
 
 using Neo.Collector.Models;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Xml;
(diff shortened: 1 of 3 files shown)


shargon (dismissed):
Good change!

ajara87 (approved):

Wi1l-B0t (approved):

shargon (approved):

PR #664, opened 2026-06-23
Keep batch command text after a mid-word quote in the same token

Jim8y:
## Problem

`BatchCommand.SplitCommandLine` is adapted from the [command-line-api](https://github.com/dotnet/command-line-api) string splitter, but the adaptation collapsed the splitter's two independent state machines (word boundary and quote boundary) into one enum plus a single `skipQuoteAtIndex`.

As a result, a quote that appears **mid-word** — while a token is already started — makes the closing quote terminate the token. Text following the closing quote becomes a separate token:

| Input | Current | Reference splitter |
|-------|---------|--------------------|
| `foo"bar"baz` | `foobar`, `baz` | `foobarbaz` |
| `a"b"c` | `ab`, `c` | `abc` |

## Fix

Restore the upstream two-state machine: track word boundaries (`seeking`) and quote boundaries (`seekingQuote`) independently, and strip quotes when materializing each token. A mid-word quote now only toggles quote mode instead of ending the token.

The existing neo-express behavior of rejecting an unbalanced quote (`FormatException`) — which the upstream splitter does not do — is preserved.

## Tests

Added two cases to `BatchCommandParserTests` for the mid-word quote inputs above. They fail against the previous code and pass with the fix. The existing `SplitCommandLine_reports_unbalanced_quotes` and `SplitCommandLine_preserves_balanced_quoted_tokens` tests continue to pass. Release build is clean and `dotnet format --verify-no-changes` reports no changes.

### DIFF
--- a/src/neoxp/Commands/BatchCommand.cs
+++ b/src/neoxp/Commands/BatchCommand.cs
@@ -394,7 +394,13 @@ namespace NeoExpress.Commands
 
             string CurrentToken()
             {
-                return memory.Slice(startTokenIndex, IndexOfEndOfToken()).ToString().Replace("\"", string.Empty);
+                var slice = memory.Slice(startTokenIndex, IndexOfEndOfToken());
+                var token = slice.ToString();
+                // Mid-word quotes are stripped from the token; skip the allocation
+                // for the common case of a token that contains no quote characters.
+                return slice.Span.IndexOf('\"') >= 0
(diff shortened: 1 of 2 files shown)


ajara87 (commented):
lgtm

shargon (dismissed):

ajara87 (approved):

shargon (approved):

PR #666, opened 2026-06-23
Add a byte-array Be overload to StorageItemAssertions

Jim8y:
## Motivation

`StorageItemAssertions` can assert that a `StorageItem` holds a `BigInteger`, `UInt160`, or `UInt256` ([StorageItemAssertions.cs:40-47](https://github.com/neo-project/neo-express/blob/master/src/assertions/StorageItemAssertions.cs#L40-L47)), but there is no way to assert it holds a specific **raw byte sequence**. The shared `Be<T>` helper is constrained to `IEquatable<T>`, which a byte array does not satisfy.

Persisting an arbitrary byte blob (a serialized struct, a stored key, a hash) is one of the most common things a Neo N3 contract does, and the sibling `StackItemAssertions` already exposes byte-array equality via `BeEquivalentTo(ReadOnlySpan<byte>)` ([StackItemAssertions.cs:183](https://github.com/neo-project/neo-express/blob/master/src/assertions/StackItemAssertions.cs#L183)). This is a missing-but-expected overload, not a stylistic choice.

## Change

Add two public overloads to `StorageItemAssertions`:
- `Be(ReadOnlySpan<byte> expected, …)` — compares the stored value with `SequenceEqual` and reports a hex diff on mismatch, mirroring `StackItemAssertions`.
- `Be(byte[] expected, …)` — a convenience overload forwarding to the span form.

Purely additive; the signatures are distinct from the existing typed overloads, so no existing call resolves differently.

## Tests

Added to `StorageItemAssertionsTests`: a match case, a mismatch case (asserts the hex of the differing bytes appears in the message), and a `because` reason case. Inverting the comparison in the production code makes all three fail, confirming they exercise the real logic. Full `test-assertions` suite (21 tests) passes; `dotnet format --verify-no-changes` is clean.

### DIFF
--- a/src/assertions/StorageItemAssertions.cs
+++ b/src/assertions/StorageItemAssertions.cs
@@ -45,5 +45,21 @@ namespace Neo.Assertions
 
         public AndConstraint<StorageItemAssertions> Be(UInt256 expected, string because = "", params object[] becauseArgs)
             => Be<UInt256>(expected, bytes => new UInt256(bytes.Span), because, becauseArgs);
+
+        public AndConstraint<StorageItemAssertions> Be(ReadOnlySpan<byte> expected, string because = "", params object[] becauseArgs)
+        {
+            var span = Subject.Value.Span;
+
+            Execute.Assertion
+                .BecauseOf(because, becauseArgs)
+                .ForCondition(span.SequenceEqual(expected))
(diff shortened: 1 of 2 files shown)


shargon (dismissed):
Ok, but it's not used

Jim8y:

PR #667, opened 2026-06-23
Add event-name and negative-sender assertions to NotifyEventArgsAssertions

Jim8y:
## Motivation

`NotifyEventArgsAssertions` exposes only two assertions ([NotifyEventArgsAssertions.cs](https://github.com/neo-project/neo-express/blob/master/src/assertions/NotifyEventArgsAssertions.cs)):
- `BeEquivalentTo<T>(Expression<Action<T>>)` — requires a strongly-typed contract-method lambda **and every argument value**.
- `BeSentBy(...)` — with no negative counterpart.

Two routine test needs have no lightweight assertion today:
1. Confirm an event of a given name (e.g. `Transfer`) was emitted, without restating its arguments.
2. Confirm an event was **not** raised by a particular contract (e.g. checking a proxy/re-entrancy path did not directly emit it).

## Change

Add, consistent with the existing `BeSentBy` overloads:
- `HaveEventName(string expected, …)`
- `NotBeSentBy(UInt160 expected, …)` and `NotBeSentBy(ContractState expected, …)`

Purely additive.

## Tests

New `NotifyEventArgsAssertionsTests` (the class previously had no coverage): match / mismatch / `because`-reason for `HaveEventName`, and pass / fail for `NotBeSentBy`. Inverting both conditions in the production code makes all five fail, confirming they exercise the real logic. Full `test-assertions` suite (23 tests) passes; `dotnet format --verify-no-changes` is clean.

### DIFF
--- a/src/assertions/NotifyEventArgsAssertions.cs
+++ b/src/assertions/NotifyEventArgsAssertions.cs
@@ -58,5 +58,28 @@ namespace Neo.Assertions
 
             return new AndConstraint<NotifyEventArgsAssertions>(this);
         }
+
+        public AndConstraint<NotifyEventArgsAssertions> NotBeSentBy(ContractState expected, string because = "", params object[] becauseArgs)
+            => NotBeSentBy(expected.Hash, because, becauseArgs);
+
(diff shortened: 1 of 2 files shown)


shargon (commented):
This extension should be added when used

Jim8y:

PR #668, opened 2026-06-23
Add UInt160 and UInt256 prefix overloads to NeoStorage.StorageMap

Jim8y:
## Motivation

The `NeoStorage` navigation helpers in [Extensions.cs](https://github.com/neo-project/neo-express/blob/master/src/test-harness/Extensions.cs) are asymmetric. The single-key `TryGetValue` family has `UInt160` ([:183](https://github.com/neo-project/neo-express/blob/master/src/test-harness/Extensions.cs#L183)) and `UInt256` ([:186](https://github.com/neo-project/neo-express/blob/master/src/test-harness/Extensions.cs#L186)) overloads, but the parallel `StorageMap(...)` prefix family only has `byte`, `string`, and `ReadOnlyMemory<byte>` forms.

Compound storage keys of the form *prefix + UInt160* are everywhere in Neo contracts (NEP-17 balance maps keyed by account, allowance and ownership maps). A test author who wants one account's sub-map currently has to hand-serialize the hash (`hash.ToArray().AsMemory()`), because there is no implicit `UInt160`→`ReadOnlyMemory<byte>` conversion — even though the sibling `TryGetValue(UInt160)` already hides exactly that.

## Change

Add `StorageMap(UInt160)` and `StorageMap(UInt256)`, each delegating to the existing `ReadOnlyMemory<byte>` overload with the same serialization the `TryGetValue` siblings use. Six lines, purely additive.

## Tests

Added `StorageMap_UInt160_prefix_strips_prefix` and `StorageMap_UInt256_prefix_strips_prefix` to `NeoStorageExtensionsTests`, modeled on the existing `StorageMap_byte_*` and `TryGetValue_UInt160_*` tests: they assert matching entries are returned with the prefix stripped and non-matching entries excluded. Replacing the prefix with an empty span makes both fail, confirming they pin the stripping behavior. Full `test-harness` suite (9 tests) passes; `dotnet format --verify-no-changes` is clean.

### DIFF
--- a/src/test-harness/Extensions.cs
+++ b/src/test-harness/Extensions.cs
@@ -146,6 +146,12 @@ namespace NeoTestHarness
             => storages.Where(kvp => kvp.Key.Span.StartsWith(prefix.Span))
                 .ToDictionary(kvp => kvp.Key.Slice(prefix.Length), kvp => kvp.Value, MemorySequenceComparer.Default);
 
+        public static NeoStorage StorageMap(this NeoStorage storages, UInt160 prefix)
+            => storages.StorageMap(prefix.ToArray().AsMemory());
+
+        public static NeoStorage StorageMap(this NeoStorage storages, UInt256 prefix)
+            => storages.StorageMap(prefix.ToArray().AsMemory());
+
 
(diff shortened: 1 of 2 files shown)


shargon (commented):
It's dead code?

Jim8y:

PR #681, opened 2026-06-23
Correct the stop command multi-node error message

Jim8y:
## Problem

Running `neoxp stop` on a multi-node chain without `--node-index` or `--all` throws:

> node index or --all must be specified when **resetting** a multi-node chain

This is the `stop` command ([`[Command("stop", …)]`](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Commands/StopCommand.cs#L15)), not reset. The string was copy-pasted from [ResetCommand.cs:61](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Commands/ResetCommand.cs#L61) and names the wrong action, which is confusing when diagnosing why a stop failed.

## Fix

Correct "resetting" to "stopping", and extract the node-index resolution into an `internal static ResolveNodeIndex(int?, int)` helper (mirroring the existing `FastForwardCommand.ValidateCount` pattern) so it can be unit-tested without a running chain. Behavior is otherwise unchanged.

## Tests

`StopCommandTests` covers the corrected message for a multi-node chain, the single-node default (index 0), an in-range index, and an out-of-range rejection. Restoring the "resetting" wording makes the message test fail. Release build is clean and `dotnet format --verify-no-changes` reports no changes.

### DIFF
--- a/src/neoxp/Commands/StopCommand.cs
+++ b/src/neoxp/Commands/StopCommand.cs
@@ -53,20 +53,27 @@ namespace NeoExpress.Commands
             }
             else
             {
-                var nodeIndex = NodeIndex.HasValue
-                    ? NodeIndex.Value
-                    : chain.ConsensusNodes.Count == 1
-                        ? 0
-                        : throw new InvalidOperationException("node index or --all must be specified when resetting a multi-node chain");
-
-                if (nodeIndex < 0 || nodeIndex >= chain.ConsensusNodes.Count)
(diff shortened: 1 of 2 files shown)


ajara87 (approved):

shargon (commented):
What changed?

ajara87:
> What changed?

refactor and UTs

shargon (approved):

PR #682, opened 2026-06-23
Dispose the RpcClient owned by OnlineNode

Jim8y:
## Problem

`OnlineNode` is an `IExpressNode` (which is `IDisposable`) that constructs and holds an `RpcClient` ([OnlineNode.cs:34](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/OnlineNode.cs#L34), [:43](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/OnlineNode.cs#L43)), but its `Dispose()` body is empty ([:47](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/OnlineNode.cs#L47)). `RpcClient` is `IDisposable` and owns an `HttpClient` that it releases in its own `Dispose`.

Every neoxp command that runs against a live node does `using var expressNode = chainManager.GetExpressNode()`, and `GetExpressNode()` returns a new `OnlineNode` when connected — so each such invocation leaks the `HttpClient` and its socket handle. The sibling `OfflineNode` disposes its resources, and `TransactionExecutor` uses `using var rpcClient = new RpcClient(uri)`, confirming the codebase's own convention.

## Fix

```csharp
public void Dispose()
{
    rpcClient.Dispose();
}
```

## Tests

`OnlineNodeTests` constructs an `OnlineNode` (its constructor performs no network I/O), disposes it, and asserts the owned `RpcClient` now throws `ObjectDisposedException` on use. Against the previous empty `Dispose()` the client is still live and instead attempts a network call (`HttpRequestException`), so the test fails. Release build is clean and `dotnet format --verify-no-changes` reports no changes.

### DIFF
--- a/src/neoxp/Node/OnlineNode.cs
+++ b/src/neoxp/Node/OnlineNode.cs
@@ -33,6 +33,7 @@ namespace NeoExpress.Node
         readonly ExpressChain chain;
         readonly RpcClient rpcClient;
         readonly Lazy<KeyPair[]> consensusNodesKeys;
+        bool disposedValue;
 
         public ProtocolSettings ProtocolSettings { get; }
 
@@ -46,7 +47,11 @@ namespace NeoExpress.Node
 
         public void Dispose()
         {
-            rpcClient.Dispose();
+            if (!disposedValue)
+            {
+                rpcClient.Dispose();
+                disposedValue = true;
+            }
         }
 
         public async Task<IExpressNode.CheckpointMode> CreateCheckpointAsync(string checkPointPath)
--- a/src/neoxp/Node/OnlineNode.cs
+++ b/src/neoxp/Node/OnlineNode.cs
@@ -49,8 +49,8 @@ namespace NeoExpress.Node
         {
             if (!disposedValue)
             {
-                rpcClient.Dispose();
                 disposedValue = true;
+                rpcClient.Dispose();
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/neoxp/Node/OnlineNode.cs
```diff
@@ -46,6 +46,7 @@ public OnlineNode(ProtocolSettings settings, ExpressChain chain, ExpressConsensu
 
         public void Dispose()
         {
+            rpcClient.Dispose();
```
ajara87: could we add the dispose pattern?

private bool _disposed;

public void Dispose()
{
    if (_disposed) return;
    _disposed = true;
    rpcClient.Dispose();
}

shargon (dismissed):

Wi1l-B0t (dismissed):

ajara87 (dismissed):

src/neoxp/Node/OnlineNode.cs
```diff
@@ -46,7 +47,11 @@ public OnlineNode(ProtocolSettings settings, ExpressChain chain, ExpressConsensu
 
         public void Dispose()
         {
-            rpcClient.Dispose();
+            if (!disposedValue)
+            {
+                rpcClient.Dispose();
+                disposedValue = true;
```
shargon, line 52: First change the flag

shargon (approved):

ajara87 (approved):

PR #683, opened 2026-06-23
Add ApplicationEngine Halt and Fault assertions to Neo.Assertions

Jim8y:
## Motivation

The most common thing a Neo contract test does is assert that the VM halted (or faulted) after running a script, but `Neo.Assertions` had no helper for `ApplicationEngine` ([NeoAssertionsExtensions.cs](https://github.com/neo-project/neo-express/blob/master/src/assertions/NeoAssertionsExtensions.cs) only covers `StackItem`, `NotifyEventArgs`, `StorageItem`). Authors fall back to:

```csharp
engine.State.Should().Be(VMState.HALT);
```

That bare comparison gives an unhelpful message on an unexpected `FAULT` — it never surfaces `engine.FaultException`, so a failing test just says "expected HALT, found FAULT" with no clue why the contract aborted.

## Change

- Add `ApplicationEngineAssertions` with `Halt(...)` and `Fault(...)`. A failed `Halt()` reports the actual state **and** the fault exception message.
- Add `Should(this ApplicationEngine)` to `NeoAssertionsExtensions`.

Purely additive.

## Tests

`ApplicationEngineAssertionsTests` initializes a genesis ledger over a `MemoryStore` and runs trivial `RET` (HALT) and `ABORT` (FAULT) scripts. It asserts `Halt()`/`Fault()` pass on the matching state and throw on the other, and that a failed `Halt()` on an aborted engine surfaces the fault exception in its message. Inverting the `Halt` condition makes all three fail, confirming they exercise the real logic. Full `test-assertions` suite (21) passes; `dotnet format --verify-no-changes` is clean.

### DIFF
--- a/src/assertions/ApplicationEngineAssertions.cs
+++ b/src/assertions/ApplicationEngineAssertions.cs
@@ -1,0 +1,51 @@
+// Copyright (C) 2015-2026 The Neo Project.
+//
+// ApplicationEngineAssertions.cs file belongs to neo-express project and is free
+// software distributed under the MIT software license, see the
+// accompanying file LICENSE in the main directory of the
+// repository or https://opensource.org/license/MIT for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using FluentAssertions;
+using FluentAssertions.Execution;
+using FluentAssertions.Primitives;
+using Neo.SmartContract;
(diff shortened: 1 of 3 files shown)


shargon (commented):
Should be added when used, otherwise it's dead code

Jim8y:

PR #689, opened 2026-06-23
Skip malformed Transfer notifications in ToolkitRpcServer.GetNep17Balances

Jim8y:
## Problem

`ToolkitRpcServer.GetNep17Balances` (behind worknet's `getnep17balances` RPC) indexes `notification.State[0]` and `notification.State[1]` ([ToolkitRpcServer.cs:191](https://github.com/neo-project/neo-express/blob/master/src/bctklib/plugins/ToolkitRpcServer.cs#L191)) **without checking `notification.State.Count`**.

These notifications come from contract-emitted `Transfer` events, whose argument count is fully controlled by the (possibly buggy or malicious) contract. A `Transfer` raised with fewer than two state items makes the `Neo.VM` array indexer throw, aborting the entire RPC call instead of skipping the bad event.

Every sibling path already guards this: `GetNep11Balances` checks `State.Count < 4`, `GetTransfers` checks the count, and neoxp's `TransferNotificationRecord.Create` starts with `State.Count < 3`. `GetNep17Balances` was the lone outlier.

## Fix

Skip a notification whose `State.Count < 3` (a well-formed NEP-17 Transfer carries from/to/amount), before indexing — matching the sibling guards.

## Tests

`ToolkitRpcServerNep17Tests` initializes a genesis ledger (so the genesis account holds the native NEP-17 NEO/GAS), then drives `GetNep17Balances` with a stub notifications provider returning a `Transfer` whose state array is empty. With the fix the query completes and returns the balances; against the previous code it throws `ArgumentOutOfRangeException`. Full `test.bctklib` suite (291) passes; `dotnet format --verify-no-changes` is clean.

### DIFF
--- a/src/bctklib/plugins/ToolkitRpcServer.cs
+++ b/src/bctklib/plugins/ToolkitRpcServer.cs
@@ -184,6 +184,11 @@ namespace Neo.BlockchainToolkit.Plugins
 
                 foreach (var (blockIndex, _, _, notification) in notifications)
                 {
+                    // a well-formed NEP-17 Transfer carries from/to/amount; skip malformed
+                    // events so a buggy or malicious contract cannot abort the whole query
+                    if (notification.State.Count < 3)
+                        continue;
+
                     // iterate backwards thru the notifications looking for all the Transfer events from a contract
                     // in assets where a Transfer event hasn't already been recorded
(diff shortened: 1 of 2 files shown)


shargon (approved):

Wi1l-B0t (approved):

ajara87:
lgtm

PR #691, opened 2026-06-23
Write the generated contract interface once and surface a persistent write failure

Jim8y:
## Problem

`NeoContractInterface.FileOperationWithRetry` ([NeoContractInterface.cs:64](https://github.com/neo-project/neo-express/blob/master/src/build-tasks/NeoContractInterface.cs#L64)) has two retry-loop defects:

```csharp
for (int retriesLeft = 6; retriesLeft > 0; retriesLeft--)
{
    try { operation(); }
    catch (IOException ex) when (… && retriesLeft > 0) { Task.Delay(100).Wait(); continue; }
}
```

1. **No break on success** — the only flow control is the `continue` in the catch, so a *successful* `operation()` falls through and runs again: the output file is written **six times** per generation.
2. **Silent failure on exhaustion** — the catch filter `retriesLeft > 0` is always true (the loop guarantees it), so when the sharing violation never clears, the loop runs all six attempts, then exits **silently** — the `IOException` is dropped, leaving no output file and no error.

## Fix

- `return;` after a successful `operation()` so the loop is only re-entered for the sharing-violation `IOException`.
- Retry only while a further attempt remains (`retriesLeft > 1`), so the **last** failure propagates instead of being swallowed.

## Tests

`TestNeoContractInterface` covers: a successful operation runs exactly once; an operation that throws one sharing violation is retried exactly twice; and an operation that always throws the sharing violation **surfaces the `IOException`** after exactly six attempts. Reverting either change fails the corresponding test. Builds for `net10.0` and `netstandard2.0`; full `test-build-tasks` suite (29) passes; format clean.

### DIFF
--- a/src/build-tasks/NeoContractInterface.cs
+++ b/src/build-tasks/NeoContractInterface.cs
@@ -61,7 +61,7 @@ namespace Neo.BuildTasks
 
         public string ContractNameOverride { get; set; } = "";
 
-        static void FileOperationWithRetry(Action operation)
+        internal static void FileOperationWithRetry(Action operation)
         {
             const int ProcessCannotAccessFileHR = unchecked((int)0x80070020);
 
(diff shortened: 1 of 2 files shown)


shargon (dismissed):
good catch

ajara87 (approved):

shargon (approved):

PR #694, opened 2026-06-23
Walk branch instructions once per line in the Cobertura writer

Jim8y:
## Problem

`ContractCoverageWriter.WriteLine` computes the per-line branch instruction set **twice**:
- [:129](https://github.com/neo-project/neo-express/blob/master/src/collector/Formats/CoberturaFormat.ContractCoverageWriter.cs#L129) `GetBranchRate(method, index, …)` internally calls `GetBranchInstructions` → `GetLineLastAddress`, walking the instruction map from the sequence point.
- [:147](https://github.com/neo-project/neo-express/blob/master/src/collector/Formats/CoberturaFormat.ContractCoverageWriter.cs#L147) the `<conditions>` loop calls `GetBranchInstructions(method, index)` **again**, repeating the same walk.

`WriteLine` runs once per sequence point per method, and every line is emitted twice per class (method-level and class-level `<lines>`), so the redundant walk multiplies across methods × lines × 2.

## Fix

Materialize `GetBranchInstructions(method, index)` once into a list, compute the rate from it via the existing `IEnumerable<(int,OpCode)>` overload of `GetBranchRate`, and iterate the same list for the `<conditions>`. One instruction-map walk per line instead of two.

## Verification

Behavior-preserving: rendering the `registrar`/`run1` fixture before and after the change produces **byte-identical** XML (only the wall-clock `timestamp` attribute differs). Added `CoberturaBranchRateTests` pinning the branch aggregate (`branches-valid/covered`), a specific `condition-coverage` rate, and the `<condition>` rendering — all of which flow through the changed call path. Full `test-collector` suite (5) passes; format clean.

### DIFF
--- a/src/collector/Formats/CoberturaFormat.ContractCoverageWriter.cs
+++ b/src/collector/Formats/CoberturaFormat.ContractCoverageWriter.cs
@@ -126,7 +126,10 @@ namespace Neo.Collector.Formats
             {
                 var sp = method.SequencePoints[index];
                 var hits = contract.HitMap.TryGetValue(sp.Address, out var value) ? value : 0;
-                var (branchCount, branchHit) = contract.InstructionMap.GetBranchRate(method, index, GetBranchHit);
+                // Walk the instruction map once; both the branch rate and the per-condition
+                // loop below use the same materialized set of branch instructions.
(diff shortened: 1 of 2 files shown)


shargon:
test failing

ajara87 (approved):

shargon (approved):

PR #695, opened 2026-06-23
Fix the version-keyed well-known contracts cache key

Jim8y:
## Problem

In `AutoComplete`, the cache key that invalidates the cached well-known contract manifests across neo-express upgrades is derived as ([autoComplete.ts:95](https://github.com/neo-project/neo-express/blob/master/extentions/neo3-visual-tracker/src/extension/autoComplete.ts#L95)):

```ts
const version = versionResult.message.trim().substring(256);
cacheKey = `wellKnownContracts_${version}`;
```

`neoxp -v` prints a short version string (e.g. `3.7.0`), far under 256 chars. `String.prototype.substring(256)` returns the substring **starting at** index 256 — the empty string for any real version output. So `version` is always `""` and `cacheKey` is always the constant `wellKnownContracts_`.

The cache is therefore never version-specific: after the bundled neoxp binary is upgraded (which happens when the extension updates), the extension keeps serving the **previous** version's genesis/well-known contract manifests for autocomplete instead of regenerating them.

## Fix

Cap the version at 256 chars with `substring(0, 256)` (the obviously-intended call, given the variable name and its use as a key suffix), extracted into a small pure helper so it is unit-testable without importing `vscode`.

## Tests

`wellKnownContractsCacheKey.test.ts` (node:test, run via a new `test:cachekey` script) asserts the key is version-specific, trims whitespace, distinguishes two versions, and caps at 256 chars. Under the old `substring(256)` all four fail (every version collapses to `wellKnownContracts_`). `npm run typecheck` passes and lint reports no new problems. The test also runs under the aggregate `npm test` added in #676.

### DIFF
--- a/extentions/neo3-visual-tracker/package.json
+++ b/extentions/neo3-visual-tracker/package.json
@@ -343,6 +343,7 @@
     "watch": "concurrently -r npm:watch-*",
     "watch-ext": "webpack --config src/extension/webpack.config.js --mode development --watch --stats-error-details",
     "watch-panel": "webpack --config src/panel/webpack.config.js --mode development --watch --stats-error-details",
+    "test:cachekey": "node --test -r ts-node/register src/extension/util/wellKnownContractsCacheKey.test.ts",
     "test:quickstart": "node --test -r ts-node/register src/panel/components/views/quickStartActions.test.ts",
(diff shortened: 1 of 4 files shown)


shargon (dismissed):

ajara87 (dismissed):

ajara87:
conflicts

ajara87 (approved):

shargon (approved):

PR #698, opened 2026-06-23
Report non-compliance and exit non-zero from contract validate

Jim8y:
## Problem

`neoxp contract validate nep11` / `nep17` ([ContractCommand.Validate.cs](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Commands/ContractCommand.Validate.cs)) printed a message **only** when the contract was compliant and **always returned exit code 0**:

```csharp
if (nep17)
    await console.Out.WriteLineAsync($"{scriptHash} is NEP-17 compliant.");
return 0;
```

So a script running `neoxp contract validate nep17 $HASH` against a **non-compliant** contract gets empty stdout and a **success** exit code — the exact opposite of what a validate command should report.

Separately, the positional argument is described as `"Path to contract .nef file"` but is parsed as a contract script hash (`UInt160.TryParse(ContractHash)`).

## Fix

- Print an explicit `"… is NOT <standard> compliant."` message and return a non-zero exit code in the non-compliant case, via a small testable `ComplianceResult` helper shared by both subcommands.
- Correct the argument description to `"Contract script hash"`.

## Tests

`ContractValidateTests` covers both outcomes of `ComplianceResult`: compliant → the "is … compliant." message and exit 0; non-compliant → the "is NOT … compliant." message and a non-zero exit. Reverting the non-compliant branch to exit 0 fails the test. Release build is clean and `dotnet format --verify-no-changes` reports no changes.

### DIFF
--- a/src/neoxp/Commands/ContractCommand.Validate.cs
+++ b/src/neoxp/Commands/ContractCommand.Validate.cs
@@ -29,6 +29,11 @@ internal partial class ContractCommand
                 ? ($"{scriptHash} is {standard} compliant.", 0)
                 : ($"{scriptHash} is NOT {standard} compliant.", 1);
 
+        internal static Task WriteMessageAsync(IConsole console, string message)
+            => string.IsNullOrWhiteSpace(message)
+                ? Task.CompletedTask
+                : console.Out.WriteLineAsync(message);
+
         [Command("nep11", Description = "Checks if contract is NEP-11 compliant")]
         public class Nep11Compliant
         {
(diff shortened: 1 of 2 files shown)


ajara87 (dismissed):

src/neoxp/Commands/ContractCommand.Validate.cs
```diff
@@ -50,10 +57,9 @@ internal async Task<int> OnExecuteAsync(CommandLineApplication app, IConsole con
                     using var expressNode = chainManager.GetExpressNode();
                     var nep11 = await expressNode.IsNep11CompliantAsync(scriptHash).ConfigureAwait(false);
 
-                    if (nep11)
-                        await console.Out.WriteLineAsync($"{scriptHash} is NEP-11 compliant.");
-
-                    return 0;
+                    var (message, exitCode) = ComplianceResult(scriptHash, "NEP-11", nep11);
+                    await console.Out.WriteLineAsync(message);
```
shargon: If message is empty it print an empty line, is that expected?

shargon (approved):

ajara87 (approved):

PR #699, opened 2026-06-23
Use a reference set for StackItem.GetSize cycle detection

Jim8y:
## Problem

`Extensions.GetSize` tracks already-visited compound items in a `List<CompoundType>` and checks membership with an O(n) scan **per node** ([Extensions.cs:327](https://github.com/neo-project/neo-express/blob/master/src/bctklib/Extensions.cs#L327), [:335](https://github.com/neo-project/neo-express/blob/master/src/bctklib/Extensions.cs#L335)):

```csharp
if (serialized.Any(p => ReferenceEquals(p, array)))
    throw new NotSupportedException();
serialized.Add(array);
```

For a StackItem graph with *n* compound nodes this is **O(n²)**. `GetSize` runs on the `NotificationRecord.Size` serialization path (every persisted notification), so a large or deeply nested notification state pays the quadratic cost.

## Fix

Track visited items in a `HashSet<CompoundType>` keyed by `ReferenceEqualityComparer`, using `!serialized.Add(item)` for the cycle check — O(1) membership. Cycle detection (by reference) and the computed size are unchanged.

## Tests

`StackItemGetSizeTests` pins the computed size of a nested array/byte-string/integer structure (identical to the previous logic) and confirms a self-referential array still throws `NotSupportedException`. Both hold on master and with the change (behavior-preserving). Full `test.bctklib` suite (313) passes; format clean.

### DIFF
--- a/src/bctklib/Extensions.cs
+++ b/src/bctklib/Extensions.cs
@@ -300,7 +300,8 @@ namespace Neo.BlockchainToolkit
         {
             maxSize ??= ExecutionEngineLimits.Default.MaxItemSize;
             int size = 0;
-            var serialized = new List<VM.Types.CompoundType>();
+            // reference set for cycle detection: O(1) membership instead of an O(n) scan per node
+            var serialized = new HashSet<VM.Types.CompoundType>(ReferenceEqualityComparer.Instance);
             var unserialized = new Stack<StackItem>();
             unserialized.Push(item);
             while (unserialized.Count > 0)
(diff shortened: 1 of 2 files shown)


shargon (dismissed):

src/bctklib/Extensions.cs
```diff
@@ -324,17 +325,15 @@ public static int GetSize(this StackItem item, uint? maxSize = null)
                         }
                         break;
                     case VM.Types.Array array:
-                        if (serialized.Any(p => ReferenceEquals(p, array)))
+                        if (!serialized.Add(array))
                             throw new NotSupportedException();
```
Wi1l-B0t: No exception message?

shargon (approved):

src/bctklib/Extensions.cs
Jim8y: Added a message for the circular-reference case.

ajara87 (approved):

PR #703, opened 2026-06-23
Remove the unused positional argument from show state

Jim8y:
## Problem

`neoxp show state` declares a positional argument ([ShowCommand.State.cs:27](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Commands/ShowCommand.State.cs#L27)):

```csharp
[Argument(0, Description = "Optional block hash or index. Show most recent block if unspecified")]
internal string BlockHash { get; init; } = string.Empty;
```

but `OnExecuteAsync` only ever calls `GetLatestBlockAsync()` and **never reads `BlockHash`** — it's a copy-paste from `show block` and is silently ignored. The command reports chain-level state (block height, running status, config file), for which there is no per-block variant, and the command reference already documents `show state` with **no** argument.

So `neoxp show state <anything>` exits 0 having ignored the argument, advertising a capability that doesn't exist.

## Fix

Remove the dead argument. `show state` now takes no positional argument (matching the docs); a stray argument is rejected rather than silently ignored.

## Tests

`ShowStateCommandTests` parses the command with an extra positional and asserts it throws `CommandParsingException`. With the argument still present the positional binds to `BlockHash` and no exception is thrown, so the test fails; after removal it passes. Release build is clean and `dotnet format --verify-no-changes` reports no changes.

### DIFF
--- a/src/neoxp/Commands/ShowCommand.State.cs
+++ b/src/neoxp/Commands/ShowCommand.State.cs
@@ -24,9 +24,6 @@ namespace NeoExpress.Commands
                 this.chainManagerFactory = chainManagerFactory;
             }
 
-            [Argument(0, Description = "Optional block hash or index. Show most recent block if unspecified")]
-            internal string BlockHash { get; init; } = string.Empty;
-
             [Option(Description = "Path to neo-express data file")]
             internal string Input { get; init; } = string.Empty;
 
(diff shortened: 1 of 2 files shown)


Wi1l-B0t (approved):
Simple patch.

ajara87 (approved):

shargon (approved):

PR #704, opened 2026-06-23
Surface malformed base64 when persisting a contract storage pair

Jim8y:
## Problem

`NodeUtility.PersistStoragePair` (a local function inside `PersistStorageKeyValuePair`, [NodeUtility.cs:539](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/NodeUtility.cs#L539)) decoded the base64 key and value **inside** a `try` whose `catch` returned `false` for anything:

```csharp
try {
    snapshot.Add(new StorageKey { Id = contractId, Key = Convert.FromBase64String(storagePair.key) },
                 new StorageItem(Convert.FromBase64String(storagePair.value)));
    return true;
} catch { return false; }
```

This collapses two unrelated outcomes into the same `false`:
1. a **malformed-base64** input error (`FormatException`), and
2. the legitimate **"key already exists"** result (`DataCache.Add` throws `ArgumentException`), which the no-overwrite path uses to mean "leave the existing value".

The `expresspersiststorage` RPC forwards JSON `key`/`value` strings directly with no validation, so a bad base64 value looks like a benign no-op — and on the overwrite path the failure is ignored **after** the contract's existing storage was already cleared, reporting success while having wiped/skipped data. (The CLI path is unaffected: it re-encodes via `Convert.ToBase64String` before calling, so input is always valid there.)

## Fix

Decode the base64 **outside** the `try` so a `FormatException` surfaces, and narrow the `catch` to the `ArgumentException` that `DataCache.Add` raises for an existing key. The method is promoted to an `internal static` helper (same name/signature, so the two call sites bind unchanged) so the behavior is unit-testable.

## Tests

`PersistStoragePairTests` (MemoryStore + StoreCache, no node): a malformed base64 key throws `FormatException`; a valid pair returns `true`, and a second write to the same key returns `false` (pinning the preserved no-overwrite path). Reverting to the catch-all form makes the malformed case return `false` instead of throwing, failing the test. Release build is clean and `dotnet format --verify-no-changes` reports no changes.

### DIFF
--- a/src/neoxp/Node/NodeUtility.cs
+++ b/src/neoxp/Node/NodeUtility.cs
@@ -535,21 +535,24 @@ namespace NeoExpress.Node
             if (dirty)
                 snapshot.Commit();
             return localContract.Id;
+        }
 
-            static bool PersistStoragePair(DataCache snapshot, int contractId, (string key, string value) storagePair)
+        internal static bool PersistStoragePair(DataCache snapshot, int contractId, (string key, string value) storagePair)
+        {
+            // Decode outside the try so a malformed base64 key/value surfaces as a
+            // FormatException instead of being reported as a benign "already exists".
+            var key = Convert.FromBase64String(storagePair.key);
+            var value = Convert.FromBase64String(storagePair.value);
+            try
             {
-                try
-                {
-                    snapshot.Add(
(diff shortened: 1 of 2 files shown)


Wi1l-B0t (approved):
Minor Fix.

ajara87 (approved):

shargon (commented):
Lgtm

PR #705, opened 2026-06-23
Constrain the expresscreatecheckpoint RPC path to the working directory

Jim8y:
## Problem

`ExpressCreateCheckpoint` ([ExpressRpcServerPlugin.cs:241](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/Modules/ExpressRpcServerPlugin.cs#L241)) takes the checkpoint `filename` straight from the RPC request and passes it to `RocksDbExpressStorage.CreateCheckpoint` with **no path validation**. A caller can therefore create a checkpoint directory at an arbitrary location via path traversal (`../../…`) or an absolute path (`/tmp/…`).

The CLI path is already restricted — `ExpressChainManager.CreateCheckpointAsync` runs `ResolveCheckpointFileName`, which enforces the checkpoint stays within the current directory ([ExpressChainManager.cs:139](https://github.com/neo-project/neo-express/blob/master/src/neoxp/ExpressChainManager.cs#L139)) before sending the request. The RPC handler applied no equivalent check.

## Fix

Resolve the filename and reject any path that escapes the node's working directory before creating the checkpoint, mirroring the CLI restriction. The check is an `internal static` helper (`EnsureCheckpointPathWithinDirectory`) that rejects both directory-traversal and absolute-outside paths and returns the resolved full path. The thrown `ArgumentException` is wrapped into the existing `InvalidParams` RPC error.

## Behavior note (please review)

This mirrors the CLI's existing within-directory model, with one honest caveat: the RPC now validates against the **node's** working directory (where `neoxp run` was started). In the common case the node and the `checkpoint create` process share that directory, so legitimate checkpoints are unaffected. The only legitimate scenario affected is creating a checkpoint from a directory **outside** the node's working-directory tree, which is uncommon. The node cannot know the calling CLI process's working directory, so this is the closest safe restriction without a larger redesign of where RPC checkpoints may be written.

## Tests

`ExpressCreateCheckpointPathTests` covers the helper: a path inside the base directory is accepted and returned as a full path; a traversal path and an absolute path outside the base directory both throw `ArgumentException`. Removing the check makes the two rejection tests fail. Release build is clean and `dotnet format --verify-no-changes` reports no changes.

### DIFF
--- a/src/neoxp/Node/Modules/ExpressRpcServerPlugin.cs
+++ b/src/neoxp/Node/Modules/ExpressRpcServerPlugin.cs
@@ -249,6 +249,10 @@ namespace NeoExpress.Node
             {
                 try
                 {
+                    // The filename comes straight from the RPC request. Constrain it to the
+                    // node's working directory (the same restriction the CLI applies before
+                    // sending) so a caller cannot create a checkpoint at an arbitrary path.
+                    filename = EnsureCheckpointPathWithinDirectory(filename, System.IO.Directory.GetCurrentDirectory());
                     rocksDbExpressStorage.CreateCheckpoint(filename, neoSystem.Settings.Network, neoSystem.Settings.AddressVersion, nodeAccountAddress);
                 }
                 catch (Exception ex) when (ex is ArgumentException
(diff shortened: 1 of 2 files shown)


Wi1l-B0t (approved):
Minor Fix.

shargon (approved):

PR #707, opened 2026-06-23
Stop scanning contract storage twice per traced instruction

Jim8y:
## Problem

`StorageRecord.Write` ([StorageRecord.cs:41](https://github.com/neo-project/neo-express/blob/master/src/bctklib/trace-models/StorageRecord.cs#L41)) enumerates its `storages` argument **twice**:

```csharp
var count = storages.Count();        // enumeration 1 (to size the map header)
...
foreach (var (key, item) in storages) // enumeration 2 (to write the entries)
```

That argument is the **lazy iterator returned by `DataCache.Find`** — a prefix scan over the snapshot — passed straight through from `TraceApplicationEngine.WriteStorages`:

```csharp
var storages = SnapshotCache.Find(StorageKey.CreateSearchPrefix(contractState.Id, default));
traceDebugSink.Storages(scriptHash, storages);
```

`WriteStorages` runs on **every executed VM instruction** ([TraceApplicationEngine.cs:83](https://github.com/neo-project/neo-express/blob/master/src/bctklib/smart-contract/TraceApplicationEngine.cs#L83), via `PostExecuteInstruction`). Because the iterator is lazy, the `.Count()` and the `foreach` each trigger a full prefix scan, so a contract with *S* storage items running *N* instructions performs **2·N** scans of *S* items instead of *N*.

## Fix

Materialize the sequence once, then size and write from the materialized copy:

```csharp
var materialized = storages as IReadOnlyCollection<(StorageKey key, StorageItem item)>
    ?? storages.ToList();
```

An already-materialized collection (such as the arrays used in tests) takes the `as IReadOnlyCollection` fast path with no extra allocation; a lazy iterator is walked exactly once. The empty-input early return is preserved.

## Why this is safe

The serialized output is **byte-for-byte identical** — same map-header count, same entry order, same encoding. The external Neo Smart Contract Debugger that consumes `.neo-trace` still receives the full per-instruction storage state; this changes only how many times the source is scanned to produce it, not what is produced.

## Tests

Added `storage_record_write_enumerates_lazy_storages_once` to `TraceDebugTests`. It wraps the storage entries in a counting lazy iterator and asserts:
- the iterator is walked **exactly once** (reverting the fix makes the count 4 instead of 2, so the test fails — verified);
- the bytes written from the lazy iterator are **identical** to those written from a pre-materialized array;
- the record still round-trips through deserialization.

Release build is clean, `dotnet format --verify-no-changes` reports no changes, and the full `TraceDebugTests` suite passes.

### DIFF
--- a/src/bctklib/trace-models/StorageRecord.cs
+++ b/src/bctklib/trace-models/StorageRecord.cs
@@ -40,8 +40,13 @@ namespace Neo.BlockchainToolkit.TraceDebug
 
         public static void Write(ref MessagePackWriter writer, MessagePackSerializerOptions options, UInt160 scriptHash, IEnumerable<(StorageKey key, StorageItem item)> storages)
         {
-            var count = storages.Count();
-            if (count <= 0)
+            // Materialize once. storages is typically a lazy iterator returned by
+            // DataCache.Find (a prefix scan over the snapshot); the previous .Count()
+            // followed by foreach enumerated it twice, scanning the contract storage
+            // twice for every VM instruction while tracing.
+            var materialized = storages as IReadOnlyCollection<(StorageKey key, StorageItem item)>
+                ?? storages.ToList();
+            if (materialized.Count <= 0)
                 return;
 
             var byteArrayFormatter = options.Resolver.GetFormatterWithVerify<byte[]>();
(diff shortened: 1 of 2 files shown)


Wi1l-B0t (approved):
Good.

shargon (approved):

Issue #708, opened 2026-06-23
Source-level (DAP) debugger for Neo smart contracts in neo-express


Jim8y:
## Summary

Neo has had no working source-level debugger since **neo-project/neo-debugger** went dormant (~2.5 years; last release v3.8.7, May 2023). neo-express is now the consolidated, actively-maintained Neo developer-tooling monorepo, and it already ships the two hardest building blocks a source-level debugger needs:

- **NEP-19 source map** — `bctklib`'s `Neo.BlockchainToolkit.Models.DebugInfo` parses `.nefdbgnfo` / `.debug.json` (documents, methods, sequence points) that `nccs` emits.
- **Execution trace capture/replay** — `bctklib/trace-models` (`TraceDebugStream`, `ITraceDebugRecord`) plus the `neotrace` tool already record per-instruction NeoVM state.

This makes neo-express the natural home for a first-party debugger and the successor to neo-debugger. It also keeps neo-devpack-dotnet lean (compiler + framework + in-process unit-test engine); the two repos meet only at the NEF / NEP-19 `.debug.json` format seam.

## Proposed architecture — trace-replay DAP adapter

A [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/) server that an editor (VS Code, etc.) drives, backed by neo-express's existing trace + source-map infrastructure:

1. **Capture** a trace of an invocation (via `neotrace` / the trace-models sink) — the per-instruction NeoVM record stream.
2. **Replay** that trace, mapping each instruction address back to a source line through `DebugInfo` sequence points, and exposing DAP requests: `setBreakpoints`, `stackTrace`, `scopes`/`variables`, `continue`, `next`/`stepIn`/`stepOut`. Because the backend is a recorded trace, this naturally supports **time-travel** (step backward), which a live in-process hook cannot.

This is deliberately *different* from a live, in-process VM hook: trace replay matches neo-express's existing model (`neotrace`), is deterministic, and reuses code already in the repo.

## Phased plan

- **[debugger 1] Source breakpoint resolution in bctklib** — line ↔ instruction-address mapping over `DebugInfo` (the source-map primitive every later phase builds on). *(PR incoming.)*
- **[debugger 2]** Trace cursor: replay an `ITraceDebugRecord` stream and expose the current instruction / source location.
- **[debugger 3]** DAP host skeleton (`initialize` / `launch` / `configurationDone`) wired to the trace cursor.
[...]

PR #711, opened 2026-06-24
[debugger 3] Add the neodebug project and trace-replay engine

Jim8y:
## What

Creates **`src/neodebug`** — the home for the first-party source-level (Debug Adapter Protocol) debugger — and lands its first execution backend:

- **`IApplicationEngine`** / **`IExecutionContext`** — the backend-agnostic seam a debug session drives. A session never knows whether it is talking to a recorded trace or a live VM.
- **`TraceReplayEngine`** — an `IApplicationEngine` backed by a recorded `.neo-trace`. It steps a `TraceDebugReader` cursor (from #710) forwards and backwards, projecting each recorded `TraceRecord` into the current engine state. Because the backend is a recording, it reports `SupportsStepBack = true` — i.e. **time-travel** (step backward) works.
- Each recorded `TraceRecord.StackFrame` is presented as an `IExecutionContext` (instruction pointer, script, evaluation stack, locals/statics/arguments).

## Design note

The engine exposes storage as **raw key/value pairs** (`GetStorages`) rather than returning a DAP presentation container, so the engine layer has **no Debug Adapter Protocol dependency** — rendering storage as debug variables is the variable/session layer's job (a later PR). The replay engine is named `TraceReplayEngine` to avoid colliding with bctklib's existing `TraceApplicationEngine`, which *writes* traces; this one *replays* them.

This is a port of the reference `neo-debugger` `adapter3` engine, adapted to current bctklib/Neo 3.10 (e.g. native-contract bytecode is no longer reachable without a snapshot, so frames whose script isn't recorded/seeded fall back to an empty script and still appear on the call stack instead of failing the replay).

## Tests

`TraceReplayEngineTests` (6 tests, all passing), driven by a hermetic in-memory trace: opens positioned at the entry step, steps forward through the recorded instructions (result stack + HALT state surface), steps **backward** for time-travel, resolves the frame script from the trace, exposes contract storage at the cursor position, and reads frame slots.

## Notes for reviewers

- Adds the new packageless library project and a `test.neodebug` xUnit project to `neo-express.sln`.
- **Stacked on #710** (`[debugger 2]`, the `TraceDebugReader` it consumes) — until that merges this PR's diff includes #710's commit. Part of the [debugger N] series tracked by #708.

Part of #708.


src/neodebug/IApplicationEngine.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// presentation-layer container, so this interface carries no Debug Adapter Protocol dependency; rendering
+    /// storage as debug variables is the session/variable layer's concern.
+    /// </remarks>
+    internal interface IApplicationEngine : IDisposable
+    {
+        event EventHandler<(UInt160 scriptHash, string scriptName, string eventName, NeoArray state)>? DebugNotify;
+        event EventHandler<(UInt160 scriptHash, string scriptName, string message)>? DebugLog;
+
+        bool CatchBlockOnStack();
+
```
shargon, line 40:
```suggestion
```

test/test.neodebug/test.neodebug.csproj
```diff
@@ -0,0 +1,26 @@
...
+        <PackageReference Include="FluentAssertions" Version="7.2.0" />
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.0.0" />
+        <PackageReference Include="xunit.runner.visualstudio" Version="3.1.5">
+          <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+          <PrivateAssets>all</PrivateAssets>
+        </PackageReference>
+        <PackageReference Include="xunit.v3" Version="3.1.0" />
+    </ItemGroup>
+    <ItemGroup>
+      <PackageReference Update="Nerdbank.GitVersioning" Version="3.8.118" />
```
shargon: not needed

shargon (commented):
I miss more summaries

Jim8y:
Addressed. Removed the unnecessary package update and added the missing summaries.

ajara87:
conflicts

src/neodebug/TraceReplayEngine.cs
```diff
@@ -0,0 +1,197 @@
...
+                return;
+            _disposed = true;
+            _reader.Dispose();
+        }
+
+        /// <inheritdoc />
+        public bool SupportsStepBack => true;
+
+        /// <inheritdoc />
+        public bool ExecuteNextInstruction()
```
shargon: Share logic with ExecutePrevInstruction() changing stepback value

src/neodebug/TraceReplayEngine.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// <summary>
+        /// Initializes a replay engine over a trace reader and positions it at the first recorded VM step.
+        /// </summary>
+        public TraceReplayEngine(TraceDebugReader reader, IEnumerable<KeyValuePair<UInt160, Script>>? seedContracts = null)
+        {
+            _reader = reader ?? throw new ArgumentNullException(nameof(reader));
+            _seedContracts = seedContracts is null ? new() : new(seedContracts);
+
+            // Advance to the first recorded VM step so the engine opens positioned at the entry point.
+            while (_reader.TryGetNext(out var record))
```
shargon, line 48: Call ExecuteNextInstruction()

src/neodebug/TraceReplayEngine.cs
Jim8y: Addressed: both step directions now share `ExecuteInstruction(stepBack)` with a single record-selection helper.

src/neodebug/TraceReplayEngine.cs
Jim8y: Addressed: `ExecuteNextInstruction` now delegates to `ExecuteInstruction(stepBack: false)`.

Jim8y:
Updated.

shargon (approved):

Wi1l-B0t (approved):

ajara87 (approved):

PR #712, opened 2026-06-24
[debugger 4] Add the DAP adapter and debug-session contract

Jim8y:
## What

Adds the Debug Adapter Protocol surface for the `neodebug` debugger:

- **`IDebugSession`** — the engine-facing contract every DAP request delegates to (threads, stack frames, scopes/variables, breakpoints, stepping, evaluate, exception info, debug-view toggle).
- **`DebugAdapter`** (`DebugAdapterBase`) — advertises capabilities (`SupportsEvaluateForHovers`, `SupportsExceptionInfoRequest`, caught/uncaught exception-breakpoint filters) and maps every DAP request onto `IDebugSession`.
- **`DebugView`** — the source ↔ disassembly presentation toggle.

## Design note

The session is constructed on launch by an injected **`DebugSessionFactory`** delegate, not a hard-wired call into a config parser. This keeps the adapter independent of *how* a session is built — trace-replay vs. live engine, and the launch.json parsing that selects between them (a later PR). It also makes the whole DAP layer unit-testable against a fake session, with no VM or trace required.

(The reference `neo-debugger` adapter called its `LaunchConfigParser` directly; injecting the factory is the one intentional deviation. The stdio host `Program` that supplies the real factory lands with that parser in a later PR, so `neodebug` stays a library here.)

## Tests

`DebugAdapterLifecycleTests` (4 tests, all passing): capabilities are advertised correctly; requests before launch throw `ProtocolException`; launch builds and starts the session via the factory; and requests (threads, continue, next, stackTrace) delegate to the session.

**Stacked on #711** (`[debugger 3]`). Part of the [debugger N] series tracked by #708.

Part of #708.


ajara87:
conflicts

src/neodebug/TraceReplayEngine.cs
```diff
@@ -0,0 +1,197 @@
...
+                return;
+            _disposed = true;
+            _reader.Dispose();
+        }
+
+        /// <inheritdoc />
+        public bool SupportsStepBack => true;
+
+        /// <inheritdoc />
+        public bool ExecuteNextInstruction()
```
shargon: Can reuseExecutePrevInstruction()  changing only stepback argument

src/neodebug/TraceReplayEngine.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// <summary>
+        /// Initializes a replay engine over a trace reader and positions it at the first recorded VM step.
+        /// </summary>
+        public TraceReplayEngine(TraceDebugReader reader, IEnumerable<KeyValuePair<UInt160, Script>>? seedContracts = null)
+        {
+            _reader = reader ?? throw new ArgumentNullException(nameof(reader));
+            _seedContracts = seedContracts is null ? new() : new(seedContracts);
+
+            // Advance to the first recorded VM step so the engine opens positioned at the entry point.
+            while (_reader.TryGetNext(out var record))
```
shargon, line 48: Call ExecutePrevInstruction()

src/neodebug/neodebug.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <ItemGroup>
+        <PackageReference Include="Microsoft.VisualStudio.Shared.VSCodeDebugProtocol" Version="17.14.10225.1" />
+    </ItemGroup>
+
+    <ItemGroup>
+        <InternalsVisibleTo Include="test.neodebug" />
+    </ItemGroup>
+
+    <ItemGroup>
+        <PackageReference Update="Nerdbank.GitVersioning" Version="3.8.118" />
```
shargon: Not required

src/neodebug/TraceReplayEngine.cs
Jim8y: Addressed in the base debugger branch: `ExecuteNextInstruction` and `ExecutePrevInstruction` now delegate to `ExecuteInstruction(stepBack: ...)`.

src/neodebug/TraceReplayEngine.cs
Jim8y: Addressed in the base debugger branch: the public step methods now delegate through the shared instruction executor.

src/neodebug/neodebug.csproj
Jim8y: Addressed: removed the redundant `Nerdbank.GitVersioning` package update from `neodebug.csproj`.

src/neodebug/DebugAdapter.cs
```diff
@@ -0,0 +1,219 @@
...
+                {
+                    new ExceptionBreakpointsFilter(CAUGHT_EXCEPTION_FILTER, "Caught Exceptions") { Default = false },
+                    new ExceptionBreakpointsFilter(UNCAUGHT_EXCEPTION_FILTER, "Uncaught Exceptions") { Default = true },
+                },
+            };
+        }
+
+        protected override void HandleLaunchRequestAsync(IRequestResponder<LaunchArguments> responder)
+        {
+            if (session != null)
```
ajara87:
```suggestion
            if (Interlocked.CompareExchange(ref _launching, 1, 0) != 0)
```
The session != null check and the session = await sessionFactory(...) assignment are not synchronized. If two launch requests arrive concurrently both can pass the null check before either completes resulting in two sessions being started

src/neodebug/DebugAdapter.cs
```diff
@@ -0,0 +1,219 @@
...
+
+            _ = LaunchAsync(responder.Arguments).ContinueWith(t =>
+            {
+                if (t.IsCompletedSuccessfully)
+                {
+                    responder.SetResponse(new LaunchResponse());
+                }
+                else
+                {
+                    var exception = t.Exception is AggregateException aggregate && aggregate.InnerExceptions.Count == 1
```
ajara87, line 100:
```suggestion
                {
                    Interlocked.Exchange(ref _launching, 0);
                    var exception = t.Exception is AggregateException aggregate && aggregate.InnerExceptions.Count == 1
```

src/neodebug/DebugAdapter.cs
```diff
@@ -0,0 +1,219 @@
...
+        private class DebugViewArguments : DebugRequestArguments
+        {
+            [Newtonsoft.Json.JsonProperty("debugView")]
+            public string DebugView { get; set; } = string.Empty;
+        }
+
+        private readonly Action<LogCategory, string> logger;
+        private readonly DebugView defaultDebugView;
+        private readonly DebugSessionFactory sessionFactory;
+        private IDebugSession? session;
```
ajara87, line 50:
```suggestion
        private IDebugSession? session;
        private int _launching; // 0 = idle, 1 = launching or launched
```

Jim8y: Addressed.

src/neodebug/DebugAdapter.cs
Jim8y: Addressed. Failed launches reset the guard so the request can be retried.

src/neodebug/DebugAdapter.cs
Jim8y: Addressed. Launch now uses an atomic guard, and I added a concurrent launch regression test.

Jim8y:
Updated.

shargon:
Conflicts

Jim8y:

shargon (approved):

ajara87 (approved):

PR #713, opened 2026-06-24
[debugger 5] Add the variable and scope rendering pipeline

Jim8y:
## What

Projects NeoVM `StackItem`s and slot variables into Debug Adapter Protocol `Variable`s and `Scope`s — the value-presentation layer the debug session renders at every stop:

- **`Extensions.ToVariable`** — typed conversion (using the debug-info `ContractParameterType`: hash160/256, public key, integer, string, byte array…) and a typeless fallback.
- **`VariableManager`** — the handle table that assigns each container its DAP `variablesReference`.
- **Containers** — lazily expanding `NeoArrayContainer` (array/struct), `NeoMapContainer`, `ByteArrayContainer` (per-byte), `SlotContainer` (eval stack / args / locals / statics), `ExecutionContextContainer` (named args/locals/statics from debug info), `EngineContainer` (gas), and `StorageContainerBase` + `StorageContainer` (storage rows, with `#storage[..]` evaluate support).

## Bug fixes carried during the port

Three latent bugs in the reference renderer are fixed rather than reproduced:
1. **Integers were mislabeled** with `Type = "Boolean"` → now `"Integer"`.
2. **Maps threw** `NotImplementedException` on any non-(bool/bytestring/integer) key → now render a hex fallback instead of failing the whole map.
3. **The variable handle table keyed on `GetHashCode`** (collisions threw a bare `Exception`, and a 0 hash collides with "no children") → now a monotonically increasing, non-zero id, reset on `Clear`.

## Drift

`ToHexString` moved to `Neo.Extensions` in Neo 3.10 (updated usings).

## Tests

`VariableRenderingTests` (8 tests, all passing): integer/boolean/null/pointer types, byte-string per-byte expansion, array expansion, mixed-key map rendering, the variable-manager id/clear contract, storage rows expanding to key/item, and named locals resolved from debug info.

The disassembly view lands with the debug session that consumes it. **Stacked on #712** (`[debugger 4]`). Part of #708.

### DIFF
--- a/src/neodebug/Extensions.cs
+++ b/src/neodebug/Extensions.cs
@@ -120,6 +120,17 @@ namespace NeoDebug.Neo3
 
         public static int GetSequenceHashCode(this byte[] array) => GetSequenceHashCode(array.AsSpan());
 
+        public static string ToMapKeyString(this Neo.VM.Types.PrimitiveType item)
+        {
+            return item switch
+            {
+                Neo.VM.Types.Boolean @bool => @bool.GetBoolean().ToString(),
+                Neo.VM.Types.ByteString byteString => byteString.GetSpan().ToHexString(),
+                Neo.VM.Types.Integer @int => @int.GetInteger().ToString(),
+                _ => item.GetSpan().ToHexString(),
+            };
+        }
+
         public static JToken ToJson(this StackItem item)
         {
             return item switch
--- a/src/neodebug/Extensions.cs
+++ b/src/neodebug/Extensions.cs
@@ -104,22 +104,6 @@ namespace NeoDebug.Neo3
             return false;
         }
 
-        // https://stackoverflow.com/a/1646913 — a deterministic hash over the bytes, used to address storage rows.
-        public static int GetSequenceHashCode(this ReadOnlySpan<byte> span)
-        {
-            unchecked
-            {
-                int hash = 17;
-                for (int i = 0; i < span.Length; i++)
-                {
-                    hash = (hash * 31) + span[i];
-                }
-                return hash;
-            }
-        }
-
-        public static int GetSequenceHashCode(this byte[] array) => GetSequenceHashCode(array.AsSpan());
-
         public static string ToMapKeyString(this Neo.VM.Types.PrimitiveType item)
         {
             return item switch
--- a/src/neodebug/Extensions.cs
+++ b/src/neodebug/Extensions.cs
@@ -1,0 +1,226 @@
+// Copyright (C) 2015-2026 The Neo Project.
+//
+// Extensions.cs file belongs to neo-express project and is free
+// software distributed under the MIT software license, see the
+// accompanying file LICENSE in the main directory of the
+// repository or https://opensource.org/license/MIT for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using Microsoft.VisualStudio.Shared.VSCodeDebugProtocol.Messages;
+using Neo;
+using Neo.BlockchainToolkit.Models;
+using Neo.Cryptography.ECC;
+using Neo.Extensions;
+using Neo.SmartContract;
+using Newtonsoft.Json.Linq;
+using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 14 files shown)


src/neodebug/VariableContainers/StorageContainerBase.cs
```diff
@@ -0,0 +1,131 @@
...
+                {
+                    return (storage.item.Value, remain.Slice(4));
+                }
+            }
+
+            throw new InvalidOperationException("Invalid storage evaluation");
+
+            static bool TryGetKeyHash(ReadOnlyMemory<char> expression, out int value)
+            {
+                if (expression.Length >= 18
```
ajara87:
```suggestion
                if (expression.Length >= 19
```
`TryGetKeyHash` should require at least 19 characters before reading `expression.Span[18]`. An input like "#storage[00000000]" has length 18 passes the current `>= 18` check and then indexes past the end instead of failing cleanly as an invalid storage expression

Jim8y: Fixed. The parser now requires at least 19 characters before reading index 18, and I added a regression test for the short storage expression.

Jim8y:
Updated. CI is green and the review thread is resolved.

shargon:
Conflicts

Jim8y:

Jim8y:
Conflicts are resolved.

src/neodebug/VariableContainers/ExecutionContextContainer.cs
```diff
@@ -0,0 +1,63 @@
...
+            var args = EnumerateSlot(manager, DebugSession.ARG_SLOTS_PREFIX, _context.Arguments, method?.Parameters);
+            var locals = EnumerateSlot(manager, DebugSession.LOCAL_SLOTS_PREFIX, _context.LocalVariables, method?.Variables);
+            var statics = EnumerateSlot(manager, DebugSession.STATIC_SLOTS_PREFIX, _context.StaticFields, _debugInfo?.StaticVariables);
+
+            return args.Concat(locals).Concat(statics);
+
+            static IEnumerable<Variable> EnumerateSlot(IVariableManager manager, string prefix, IReadOnlyList<StackItem>? slot, IReadOnlyList<DebugInfo.SlotVariable>? variableList = null)
+            {
+                variableList ??= Array.Empty<DebugInfo.SlotVariable>();
+                slot ??= Array.Empty<StackItem>();
```
Wi1l-B0t:
```suggestion
                variableList ??= [];
                slot ??= [];
```

ajara87 (dismissed):

Wi1l-B0t (dismissed):

Jim8y:
@shargon Conflicts are resolved and CI is green. Could you please confirm?

src/neodebug/Extensions.cs
```diff
@@ -0,0 +1,226 @@
...
+                }
+                return null;
+            }
+        }
+
+        public static Variable ToVariable(this StackItem item, IVariableManager manager, string name)
+        {
+            return item switch
+            {
+                Neo.VM.Types.Array array => NeoArrayContainer.Create(manager, array, name),
```
shargon, line 202: Struct and array has no difference?

src/neodebug/VariableContainers/NeoMapContainer.cs
```diff
@@ -0,0 +1,58 @@
...
+            };
+        }
+
+        public IEnumerable<Variable> Enumerate(IVariableManager manager)
+        {
+            foreach (var key in _map.Keys)
+            {
+                // Map keys are always primitive types; render the common ones, and fall back to a hex dump
+                // of the key bytes for anything else rather than failing the whole map.
+                var keyString = key switch
```
shargon: Why is not used ToVariable method?

Jim8y: Updated. Map values now use `ToVariable`; only the child name uses `ToMapKeyString`.

Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
LGTM

PR #714, opened 2026-06-24
[debugger 6] Add breakpoint, disassembly and cast collaborators

Jim8y:
## What

The three collaborators the debug session composes, landed on their own so each is independently reviewable and unit-tested before the (large) session class that ties them together:

- **`BreakpointManager`** — resolves source-line breakpoints to NeoVM instruction addresses (via debug-info sequence points, or a contract's disassembly line map when the source is named by script hash), and checks the running instruction pointer against the resolved set.
- **`DisassemblyManager`** — renders a contract's bytecode as annotated disassembly text and builds the line↔address maps that drive the disassembly debug view; cached per script and handed to the editor via a source reference.
- **`CastOperation`** — the evaluate/return-value cast kinds (`int`, `bool`, `hex`, `addr`, …).

## Drift

`DisassemblyManager` reuses bctklib's `Script`/`Instruction` extensions (`EnumerateInstructions`, `GetInstructionAddressPadding`, `GetOperandString`, `GetComment`) and `ApplicationEngine.Services` for syscall names — all present in current bctklib/Neo 3.10.

## Tests

`DebugCollaboratorsTests` (3 tests, all passing): disassembly address/line maps round-trip across all instructions; `SetBreakpoints` verifies only lines that carry a sequence point; and `CheckBreakpoint` matches the resolved instruction address (and ignores a null pointer).

**Stacked on #713** (`[debugger 5]`). The `DebugSession` keystone that composes these lands next. Part of #708.


Wi1l-B0t:
Conflcts.

Jim8y:

Jim8y:
Conflicts are resolved.

src/neodebug/Extensions.cs
```diff
@@ -0,0 +1,226 @@
...
+            return item switch
+            {
+                Neo.VM.Types.Array array => NeoArrayContainer.Create(manager, array, name),
+                Neo.VM.Types.Boolean _ => new Variable { Name = name, Value = $"{item.GetBoolean()}", Type = "Boolean" },
+                Neo.VM.Types.Buffer buffer => ByteArrayContainer.Create(manager, buffer, name),
+                Neo.VM.Types.ByteString byteString => ByteArrayContainer.Create(manager, byteString, name),
+                Neo.VM.Types.Integer @int => new Variable { Name = name, Value = $"{@int.GetInteger()}", Type = "Integer" },
+                Neo.VM.Types.InteropInterface _ => new Variable { Name = name, Value = "InteropInterface" },
+                Neo.VM.Types.Map map => NeoMapContainer.Create(manager, map, name),
+                Neo.VM.Types.Null _ => new Variable { Name = name, Value = "<null>", Type = "Null" },
```
shargon: Type should be VMType instead of string

src/neodebug/VariableContainers/StorageContainerBase.cs
```diff
@@ -0,0 +1,131 @@
...
+                {
+                    return (storage.key, remain.Slice(3));
+                }
+                else if (remain.Length >= 4 && remain.Span.Slice(0, 4).SequenceEqual("item"))
+                {
+                    return (storage.item.Value, remain.Slice(4));
+                }
+            }
+
+            throw new InvalidOperationException("Invalid storage evaluation");
```
ajara87:
```suggestion
            return (null, expression);
```
`return (null, expression)` instead of exception allowing the caller decide how to handle?

Jim8y: Updated.

ajara87 (dismissed):

src/neodebug/Extensions.cs
Jim8y: Updated to use the VM stack item type.

Wi1l-B0t (dismissed):

ajara87 (dismissed):

ajara87 (dismissed):

Wi1l-B0t:
Conflicts.

ajara87 (approved):

Wi1l-B0t (approved):

shargon (approved):

PR #715, opened 2026-06-24
[debugger 7] Add the DebugSession — end-to-end replay debugging

Jim8y:
## What

The keystone: **`DebugSession`** composes the engine seam (#711), breakpoints/disassembly (#714), and variable rendering (#713) into a working source-level debugger. It implements `IDebugSession`:

- **Stepping** — step in/out/over, continue, and (for the trace-replay engine) **step back** and reverse-continue.
- **Stack frames** — source frames (sequence-point → file/line/column) or disassembly frames (address → disassembly line), per debug view.
- **Scopes & variables** — named args/locals/statics from debug info (or raw slots in disassembly view), storage, and engine gas.
- **Breakpoints** — source breakpoints via `BreakpointManager`; **exception breakpoints** by intercepting `THROW` and applying the caught/uncaught filters.
- **Evaluate** — cast-aware expression evaluation (`#local0`, named vars, `#storage[..]`, `(int)`/`(addr)`/… casts), never throwing.
- Runtime `Notify`/`Log` surface as output events; HALT/FAULT terminate with gas and return/fault output.

## Adaptations from the reference

- Storage is wrapped from the engine's raw rows (`new StorageContainer(engine.GetStorages(..))`) since the engine layer is DAP-free.
- The exception-filter ids are shared with `DebugAdapter`; the slot/storage prefix constants come from the companion partial added in #713.
- Drift: `ToAddress` (`Neo.Wallets`) and `ToHexString` (`Neo.Extensions`) namespaces updated for Neo 3.10; `Script` aliased (DAP also defines one).

## Tests

`DebugSessionReplayTests` (5 tests) drive a **real session over a synthetic `.neo-trace`** — no compiler needed: start stops at a source sequence point; a source breakpoint stops `continue` at the right line; scopes expose the named local `x = 42` from debug info; and **step-back time-travels** to the previous line. Full `test.neodebug` suite: 26 green.

This makes `neodebug` a working replay debugger end-to-end. The live (in-process) engine, the launch.json parser + installable tool, and editor glue follow. **Stacked on #714** (`[debugger 6]`). Part of #708.

### DIFF
--- a/src/neodebug/DebugAdapter.cs
+++ b/src/neodebug/DebugAdapter.cs
@@ -49,6 +49,7 @@ namespace NeoDebug.Neo3
         private readonly DebugSessionFactory sessionFactory;
         private IDebugSession? session;
         private int launching; // 0 = idle, 1 = launching or launched
+        private int configured;
 
         public DebugAdapter(Stream @in,
                             Stream @out,
@@ -77,6 +78,7 @@ namespace NeoDebug.Neo3
         {
             return new InitializeResponse()
             {
+                SupportsConfigurationDoneRequest = true,
                 SupportsEvaluateForHovers = true,
                 SupportsExceptionInfoRequest = true,
                 ExceptionBreakpointFilters = new List<ExceptionBreakpointsFilter>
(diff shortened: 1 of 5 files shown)


Wi1l-B0t:
Conflicts.

Jim8y:

Jim8y:
Conflicts are resolved.

Wi1l-B0t:
Conflicts.

ajara87 (approved):

shargon (approved):

PR #716, opened 2026-06-24
[debugger 8] Add the live in-process debug engine

Jim8y:
## What

**`DebugApplicationEngine`** is the second `IApplicationEngine` backend (alongside the trace-replay engine from #711). It debugs a contract by stepping a **live** `TestApplicationEngine` one instruction at a time (`ExecuteNext`), so the *same* `DebugSession` and DAP adapter can debug a freshly-executed invocation — not just a recorded trace.

- Exposes the live VM's invocation stack, result stack, contract storage (from the snapshot), and runtime `Notify`/`Log` as debug events.
- Reports `SupportsStepBack = false` (`ExecutePrevInstruction` throws) — time-travel remains the trace-replay engine's domain.

## Neo 3.10 drift handled

- `SnapshotCache` property (the `Snapshot` property was removed).
- `protected override void Dispose(bool)` (the VM's `Dispose()` isn't virtual).
- `OnNotify`/`OnLog` take an `ApplicationEngine` sender (not `object`).
- storage `Find(prefix, SeekDirection.Forward)`.
- `ExecutionContext` disambiguated from `System.Threading.ExecutionContext`.
- Confirmed `ExecuteNext()` is reachable from a `TestApplicationEngine` subclass in 3.10.

## Tests

`DebugApplicationEngineTests` (3 tests): step a live `PUSH1 PUSH2 ADD RET` script to `HALT` with result `3`; backward stepping is unsupported; and an unknown hash yields no contract/storage. Full `test.neodebug` suite: 29 green.

> Note: the test initializes the genesis ledger directly because bctklib's `EnsureLedgerInitialized` is currently a stub that doesn't persist anything — flagged separately.

**Stacked on #715** (`[debugger 7]`). Part of #708.


Wi1l-B0t:
Coflicts.

Wi1l-B0t:
Coflicts.

ajara87:
ping @Jim8y

shargon (approved):

ajara87 (approved):

PR #719, opened 2026-06-24
Fix EnsureLedgerInitialized to actually persist the genesis block

Jim8y:
## Problem

`Neo.BlockchainToolkit.Extensions.EnsureLedgerInitialized(this IStore store, ProtocolSettings settings)` created the genesis block but then **skipped persisting it** — its own comment said it would "skip the engine execution" as "a simplified version." As a result the ledger was never initialized, and **any `ApplicationEngine` constructed over the resulting store threw `KeyNotFoundException`** when it read `LedgerContract.CurrentIndex` in its constructor.

The one in-tree caller, `runner/Program.cs`, relies on this to initialize the ledger for a checkpoint-backed invocation, so it was relying on a no-op.

## Fix

Persist the genesis block the way the blockchain does for block 0: over a `StoreCache` (which wraps the `IStoreSnapshot` the old comment claimed couldn't be used), run the native **OnPersist** and **PostPersist** scripts, then `Commit()`. The existing "already initialized" check keeps the call idempotent. This mirrors the working logic in `test/test.bctklib/DeployedContractFixture.cs`.

## Tests

New `EnsureLedgerInitializedTests`:
- after the call, an `ApplicationEngine` constructs and `LedgerContract.CurrentIndex` reads as `0` (previously threw);
- a second call is a no-op.

Full `test.bctklib` suite is green (**327 passed**).

### DIFF
--- a/src/bctklib/Extensions.cs
+++ b/src/bctklib/Extensions.cs
@@ -274,21 +274,39 @@ namespace Neo.BlockchainToolkit
         // replicated logic from Blockchain.OnInitialized + Blockchain.Persist
         public static void EnsureLedgerInitialized(this IStore store, ProtocolSettings settings)
         {
-            using var storeSnapshot = store.GetSnapshot();
-            if (LedgerInitialized(storeSnapshot))
+            using var snapshot = new StoreCache(store.GetSnapshot());
+            if (LedgerInitialized(snapshot))
                 return;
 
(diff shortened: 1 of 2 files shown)


shargon (approved):

Wi1l-B0t (approved):
Minor Fix.

ajara87 (approved):

PR #718, opened 2026-06-24
[debugger 10] Document NeoDebug and its launch configuration

Jim8y:
## What

Documentation for the new debugger, completing the series:

- **`docs/debugger-command-reference.md`** — installing the `Neo.Debug` global tool, the *compile → capture-trace → configure → debug* workflow, the launch-configuration properties (`program`, `invocation.trace-file`, `return-types`, `sourceFileMap`) with a sample VS Code `launch.json`, the source/disassembly views, debug-console expression evaluation (named/indexed slots, `#storage[..]`, casts), and the current trace-replay scope (with live launch noted as forthcoming).
- **`readme.md`** — adds a NeoDebug section under the tooling overview with the install command and a link to the reference.
- **`docs/trace-command-reference.md`** — points the "open the trace in…" link at the in-repo NeoDebug instead of the dormant external `neo-project/neo-debugger`.

Docs only — no code. **Stacked on #717** (`[debugger 9]`). Completes the [debugger N] series tracked by #708.

PR #720, opened 2026-06-24
[debugger 11] Wire the live engine into the launch parser

Jim8y:
## What

Completes the launcher: a launch configuration whose `invocation` specifies an **`operation`** (instead of a `trace-file`) now builds a **live** `DebugApplicationEngine` (#716) instead of reporting not-supported. So the same `DebugSession` and DAP adapter can debug a *freshly-executed* invocation, not only a recorded trace.

`LaunchConfigParser.CreateLiveEngine`:
1. stands up a fresh, single-block in-process chain (genesis `OnPersist`/`PostPersist`);
2. deploys the program NEF + sibling `.manifest.json` (the logic of `ContractManagement.Deploy`, replayed directly) under a default `CalledByEntry` signer;
3. turns `{ operation, args }` into an `EmitDynamicCall` invoke script (args via bctklib's `ContractParameterParser`);
4. builds the transaction + dummy block and positions a `DebugApplicationEngine` at the call with a permissive witness checker.

## Scope / drift

Multi-contract deployment, Neo-Express account/signer resolution, checkpoints, and oracle responses are intentionally left for later. The genesis init is inline here because bctklib's `EnsureLedgerInitialized` is a stub (fixed separately in #719). Drift: `StoreCache`, `manifest.IsValid(ExecutionEngineLimits.Default, hash)`, `EmitDynamicCall`/`ReadSerializable` in `Neo.Extensions`.

## Tests

`LiveLaunchTests` — builds a one-method contract in memory, live-launches `main()`, and runs to **HALT returning 7** (no Stderr fault). The PR9 "not yet supported" test is updated to cover an invocation with neither a `trace-file` nor an `operation`. Docs updated with the live launch configuration. Full `test.neodebug` suite: **33 green**.

**Stacked on #718** (`[debugger 10]`). Part of #708.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,6 +67,9 @@ jobs:
     - name: Push Neo.Trace Nuget.org
       run: dotnet nuget push ${{ format('./dist/Neo.Trace.{0}.nupkg', steps.nbgv.outputs.NuGetPackageVersion) }} --api-key ${{ secrets.NUGET_TOKEN }} --source https://api.nuget.org/v3/index.json
 
+    - name: Push Neo.Debug Nuget.org
+      run: dotnet nuget push ${{ format('./dist/Neo.Debug.{0}.nupkg', steps.nbgv.outputs.NuGetPackageVersion) }} --api-key ${{ secrets.NUGET_TOKEN }} --source https://api.nuget.org/v3/index.json
+
     - name: Push Blockchain-toolkit-library Nuget.org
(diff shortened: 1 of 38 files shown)


Wi1l-B0t:
Good.

Wi1l-B0t:
Conflicts.

ajara87:
ping @Jim8y

ajara87 (approved):

shargon (approved):

PR #721, opened 2026-06-24
[debugger 12] Add the VS Code extension for the Neo contract debugger

Jim8y:
## What

A VS Code extension that makes the debugger usable from the editor: it registers the **`neo-contract`** debug type and runs the `neodebug` (`Neo.Debug`) tool as a Debug Adapter Protocol host, so a Neo contract can be debugged straight from the Run and Debug view.

It contributes:
- the **debugger** (`neo-contract`) and **C# breakpoints**;
- the **launch-configuration schema** — `program`, `invocation` (either `{ trace-file }` to replay or `{ operation, args }` to run live), `return-types`, `sourceFileMap`, `debug-view`;
- **launch.json snippets** for both replay and live debugging.

The adapter executable defaults to `neodebug` on `PATH` and is overridable via the `neo-contract.debugAdapterPath` setting.

## Notes

- **Build-free**: pure JavaScript (`extension.js`), no `node_modules` and no compile step. `package.json` is valid JSON and `extension.js` passes `node --check`. Package with `npx @vscode/vsce package`.
- Independent of the `neodebug` C# branches (it references the tool by name), so this PR is **not stacked** — it pairs with the debugger tool PRs and can merge on its own.

Closes the turn-key gap noted on #708. Part of #708.


ajara87:
typo `extentions` instead of `extensions`

Jim8y:
Fixed.

ajara87 (approved):

shargon (approved):

PR #724, opened 2026-06-26
Send the required force property when updating storage on a running node

Jim8y:
## Problem

`contract storage update` fails **100% of the time against a running node**. `OnlineNode.PersistStorageKeyValueAsync` ([OnlineNode.cs:294-303](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/OnlineNode.cs#L294)) builds the `expresspersiststorage` payload with only `state` and `storage`:

```csharp
JObject o = new JObject();
o["state"] = state.ToJson();
// ... storage ...
o["storage"] = storage;
// no "force"
```

But the server handler requires `force` ([ExpressRpcServerPlugin.cs:693](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/Modules/ExpressRpcServerPlugin.cs#L693)):

```csharp
var force = ParseParam(() => Enum.Parse<ContractCommand.OverwriteForce>(RequiredStringProperty(payload, "force")));
```

`RequiredStringProperty` throws `Invalid params: missing 'force'` when the property is absent, so the request is always rejected. The sibling `PersistContractAsync` works precisely because it **does** send `force` ([OnlineNode.cs:275](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/OnlineNode.cs#L275)); only the storage path omitted it. The offline path is unaffected because `OfflineNode.PersistStorageKeyValueAsync` calls `NodeUtility.PersistStorageKeyValuePair(..., OverwriteForce.None)` directly without going through RPC.

## Fix

Send `OverwriteForce.None` — matching the offline path, which always persists storage with `OverwriteForce.None`. The payload construction is extracted into an `internal static BuildPersistStoragePayload` helper so the behavior is unit-testable.

## Tests

Added `BuildPersistStoragePayload_includes_a_force_the_server_accepts` to `OnlineNodeTests`: it builds the payload and asserts `force` is present and parses as `OverwriteForce.None` using the server's exact `Enum.Parse<OverwriteForce>` expression, plus that `state` and the `storage` entry are carried through. Removing the `force` line makes the test fail (verified). Release build is clean and `dotnet format --verify-no-changes` reports no changes.

## Note

No automated test exercised this RPC round-trip previously, which is why the regression went unnoticed.

### DIFF
--- a/test/test.workflowvalidation/OnlineNodeTests.cs
+++ b/test/test.workflowvalidation/OnlineNodeTests.cs
@@ -11,7 +11,9 @@
 using FluentAssertions;
 using Neo;
 using Neo.BlockchainToolkit.Models;
+using Neo.Json;
 using Neo.Network.RPC;
+using NeoExpress.Commands;
 using NeoExpress.Node;
 using System;
 using System.Reflection;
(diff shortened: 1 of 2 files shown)


test/test.workflowvalidation/OnlineNodeTests.cs
```diff
@@ -55,4 +57,22 @@ public void Dispose_is_idempotent()
 
         act.Should().NotThrow();
     }
+
+    [Fact]
+    public void BuildPersistStoragePayload_includes_a_force_the_server_accepts()
+    {
+        var state = new JObject();
+
```
shargon:
```suggestion
```

shargon (approved):

Wi1l-B0t (approved):

PR #725, opened 2026-06-26
Verify the NEF checksum when loading a contract for coverage

Jim8y:
## Problem

The collector's `NefFile.Load` reads the trailing 4-byte checksum but never validates it — the body is literally `// TODO: Check Checksum` ([NefFile.cs:87](https://github.com/neo-project/neo-express/blob/master/src/collector/Models/NefFile.cs#L87)):

```csharp
var checksum = reader.ReadUInt32();
// TODO: Check Checksum
return new NefFile(compiler, source, tokens, script, checksum);
```

So a corrupt or truncated `.nef` that is still structurally parseable is accepted, and `CodeCoverageCollector.LoadNef` maps coverage against a malformed contract's script with no indication anything is wrong. Upstream `Neo.SmartContract.NefFile` verifies this checksum and throws on mismatch; this reimplementation did not.

## Fix

Compute the checksum the same way the compiler writes it — the first four bytes (little-endian) of the double SHA-256 hash of every byte preceding the checksum — and reject the file when it does not match:

```csharp
var checksum = reader.ReadUInt32();
if (stream.CanSeek)
{
    var contentLength = checked((int)stream.Position - sizeof(uint));
    stream.Position = 0;
    var content = reader.ReadBytes(contentLength);
    if (ComputeChecksum(content) != checksum)
        throw new FormatException("CRC verification failed");
}
```

The verification only runs on seekable streams (the only callers open `FileStream`s). `LoadNef` already wraps parsing in a try/catch that logs ([CodeCoverageCollector.cs:135-157](https://github.com/neo-project/neo-express/blob/master/src/collector/CodeCoverageCollector.cs#L135)), so a corrupt `.nef` is now surfaced as a logged error instead of silently producing wrong coverage; valid files are unaffected.

## Tests

Added `NefFileTests` against the existing real `.nef` fixture:
- **`Load_accepts_a_real_nef_with_a_valid_checksum`** — a real compiled contract still loads, which also confirms the checksum is computed exactly as the compiler writes it (this test fails if the algorithm is wrong).
- **`Load_rejects_a_nef_whose_checksum_does_not_match`** — corrupting the trailing checksum makes `Load` throw `FormatException`. Removing the check makes this test fail (verified).

### DIFF
--- a/src/collector/Models/NefFile.cs
+++ b/src/collector/Models/NefFile.cs
@@ -101,7 +101,7 @@ namespace Neo.Collector.Models
             stream.Position = 0;
             var content = reader.ReadBytes(contentLength);
             if (ComputeChecksum(content) != checksum)
-                throw new FormatException("CRC verification failed");
+                throw new FormatException("NEF checksum verification failed");
 
             return new NefFile(compiler, source, tokens, script, checksum);
         }
--- a/src/collector/Models/NefFile.cs
+++ b/src/collector/Models/NefFile.cs
@@ -9,9 +9,11 @@
 // modifications are permitted.
 
 using System;
+using System.Buffers.Binary;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
+using System.Security.Cryptography;
 using System.Text;
 
 namespace Neo.Collector.Models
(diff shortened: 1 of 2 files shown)


Wi1l-B0t (dismissed):

src/collector/Models/NefFile.cs
```diff
@@ -84,11 +94,24 @@ public static NefFile Load(Stream stream)
...
             var checksum = reader.ReadUInt32();
-            // TODO: Check Checksum
+            // The checksum is the first four bytes of the double SHA-256 hash of
+            // every preceding byte. Verify it so a corrupt or truncated .nef is
+            // rejected instead of being silently mapped to coverage data.
+            var contentLength = checked((int)stream.Position - sizeof(uint));
+            stream.Position = 0;
+            var content = reader.ReadBytes(contentLength);
+            if (ComputeChecksum(content) != checksum)
+                throw new FormatException("CRC verification failed");
```
ajara87:
```suggestion
                throw new FormatException("NEF checksum verification failed");
```
should be NEF because of the algorithm is double sha-256?

Jim8y: Addressed.

ajara87 (approved):

Wi1l-B0t (approved):

shargon (approved):

PR #727, opened 2026-06-26
Remove an unused JObject allocation in ExpressListTokenContracts

Jim8y:
## Problem

The `expresslisttokencontracts` handler allocates a `JObject` in the loop body and then adds `contract.ToJson()` to the array instead, so the local is never used ([ExpressRpcServerPlugin.cs:147-148](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/Modules/ExpressRpcServerPlugin.cs#L147)):

```csharp
foreach (var contract in snapshot.EnumerateTokenContracts(neoSystem.Settings))
{
    var jsonContract = new JObject();   // never used
    jsonContracts.Add(contract.ToJson());
}
```

## Fix

Drop the dead allocation. The response is unchanged — `contract.ToJson()` is still what gets added — this just removes a throwaway object allocated once per token contract.


Wi1l-B0t (approved):
Simple Patch.

ajara87 (approved):

shargon (approved):

PR #728, opened 2026-06-26
Guard against a null transaction when logging a faulted block execution

Jim8y:
## Problem

`ExpressLogPlugin.OnApplicationExecuted` formats the FAULT log line with `applicationExecuted.Transaction.Hash` and **no null check** ([ExpressLogPlugin.cs:113](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/Modules/ExpressLogPlugin.cs#L113)):

```csharp
var logMessage = $"Tx FAULT: hash={applicationExecuted.Transaction.Hash}";
```

But `OnCommitting` calls it for **every** entry in `applicationExecutedList` ([ExpressLogPlugin.cs:103-106](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/Modules/ExpressLogPlugin.cs#L103)), which includes block-level `OnPersist`/`PostPersist` executions whose `Transaction` is null. A faulting block-trigger execution therefore throws a `NullReferenceException` from inside the `Blockchain.Committing` handler.

The sibling `ExpressPersistencePlugin.OnCommitting` proves `Transaction` is nullable here — it explicitly skips those entries ([ExpressPersistencePlugin.cs:140-141](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/Modules/ExpressPersistencePlugin.cs#L140)):

```csharp
if (appExec.Transaction is null)
    continue;
```

Reachability is low in practice (native `OnPersist`/`PostPersist` rarely fault), but the guard is missing.

## Fix

Format the message through a `FormatFaultLog` helper that labels a null transaction as `(block)` instead of dereferencing it (a block-level fault is worth surfacing, so it is logged rather than skipped), and null-guards the exception message for the same containerless case. Output for a normal transaction fault is unchanged.

## Tests

Added `ExpressLogPluginTests`:
- `FormatFaultLog_labels_a_block_level_fault_with_a_null_transaction` — the null-transaction case yields `Tx FAULT: hash=(block) ...` instead of throwing (reverting the guard makes this test fail, verified).
- `FormatFaultLog_uses_the_transaction_hash_when_present` — a normal fault still includes the transaction hash.
- `FormatFaultLog_returns_null_for_a_non_fault_state` — non-FAULT executions produce no log.

### DIFF
--- a/src/neoxp/Node/Modules/ExpressLogPlugin.cs
+++ b/src/neoxp/Node/Modules/ExpressLogPlugin.cs
@@ -108,15 +108,32 @@ namespace NeoExpress.Node
 
         void OnApplicationExecuted(Neo.Ledger.Blockchain.ApplicationExecuted applicationExecuted)
         {
-            if (applicationExecuted.VMState == Neo.VM.VMState.FAULT)
+            var logMessage = FormatFaultLog(
+                applicationExecuted.VMState,
+                applicationExecuted.Transaction,
+                applicationExecuted.Exception);
+            if (logMessage is not null)
             {
-                var logMessage = $"Tx FAULT: hash={applicationExecuted.Transaction.Hash}";
-                if (!string.IsNullOrEmpty(applicationExecuted.Exception.Message))
-                {
-                    logMessage += $" exception=\"{applicationExecuted.Exception.Message}\"";
-                }
                 console.Error.WriteLine($"\x1b[31m{logMessage}\x1b[0m");
             }
+        }
+
+        // Block-level executions (native OnPersist/PostPersist) have a null
+        // Transaction, so a faulting one must be labelled rather than have its
+        // hash dereferenced - the way ExpressPersistencePlugin already skips
+        // null-Transaction entries.
+        internal static string? FormatFaultLog(Neo.VM.VMState vmState, Transaction? transaction, System.Exception? exception)
+        {
+            if (vmState != Neo.VM.VMState.FAULT)
+                return null;
+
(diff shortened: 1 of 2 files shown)


Wi1l-B0t (dismissed):

ajara87 (dismissed):

src/neoxp/Node/Modules/ExpressLogPlugin.cs
```diff
@@ -108,15 +108,32 @@ void OnCommitting(NeoSystem system, Block block, DataCache snapshot, IReadOnlyLi
...
+        // Block-level executions (native OnPersist/PostPersist) have a null
+        // Transaction, so a faulting one must be labelled rather than have its
+        // hash dereferenced - the way ExpressPersistencePlugin already skips
+        // null-Transaction entries.
+        internal static string? FormatFaultLog(Neo.VM.VMState vmState, Transaction? transaction, System.Exception? exception)
+        {
+            if (vmState != Neo.VM.VMState.FAULT)
+                return null;
+
+            var hash = transaction is null ? "(block)" : transaction.Hash.ToString();
```
shargon: maybe it's not a block maybe it's better to send the ScriptContainer (maybe it's null)

Jim8y: Addressed by logging `<unknown>` for a null transaction instead of assuming it is a block or native execution.

shargon (approved):

ajara87 (approved):

PR #730, opened 2026-06-26
Parse numeric CLI arguments with the invariant culture

Jim8y:
## Problem

`decimal` options and arguments are parsed by McMaster using the host machine's **current culture**:
- `policy set <policy> <value>` ([PolicyCommand.Set.cs:37](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Commands/PolicyCommand.Set.cs#L37)) and its batch form,
- `--additional-gas` on `execute`, `contract invoke`, and `contract run` ([ExecuteCommand.cs:55](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Commands/ExecuteCommand.cs#L55), [ContractCommand.Invoke.cs:47](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Commands/ContractCommand.Invoke.cs#L47), [ContractCommand.Run.cs:53](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Commands/ContractCommand.Run.cs#L53)).

On a locale whose decimal separator is `,` (e.g. de-DE, fr-FR):
- `neoxp policy set GasPerBlock 1.5 genesis` is rejected outright with **"1.5 is not a valid value for Decimal"**;
- a value typed as `1,5` is silently read as a *different* value than on an en-US machine.

Fractional GAS-denominated values are intentional — `SetPolicyAsync` computes the decimal scale and supports up to 8 decimal places. The transfer command already avoids this by parsing its quantity with `InvariantCulture` ([`ParseQuantity`](https://github.com/neo-project/neo-express/blob/master/src/neoxp/TransactionExecutor.cs#L524), with a comment documenting exactly this bug class), but the policy/gas decimals were not given the same treatment.

## Fix

Set `ValueParsers.ParseCulture` to `CultureInfo.InvariantCulture` for **both** the main application ([Program.cs](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Program.cs#L52)) and the batch sub-application ([BatchCommand.cs](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Commands/BatchCommand.cs#L102)), via a shared `UseInvariantValueParsing` helper. Numeric CLI arguments now always use `.` as the decimal separator regardless of locale — the correct, predictable behavior for a CLI, and a generalization of the existing `ParseQuantity` handling.

## Tests

Added `CommandLineApplicationExtensionsTests.UseInvariantValueParsing_parses_a_period_decimal_under_a_comma_locale`: under `de-DE`, a `decimal` option parses `1.5` to `1.5m`. Removing the helper call makes the test fail under de-DE (verified), confirming both the bug and the fix. Release build is clean and `dotnet format --verify-no-changes` reports no changes.

### DIFF
--- a/src/neoxp/Commands/BatchCommand.cs
+++ b/src/neoxp/Commands/BatchCommand.cs
@@ -100,6 +100,7 @@ namespace NeoExpress.Commands
             for (var i = 0; i < commands.Length; i++)
             {
                 var batchApp = new CommandLineApplication<BatchFileCommands>();
+                batchApp.UseInvariantValueParsing();
                 batchApp.Conventions.UseDefaultConventions();
 
                 var args = SplitCommandLine(commands.Span[i]).ToArray();
--- a/src/neoxp/Extensions/CommandLineApplicationExtensions.cs
+++ b/src/neoxp/Extensions/CommandLineApplicationExtensions.cs
@@ -1,0 +1,28 @@
+// Copyright (C) 2015-2026 The Neo Project.
+//
+// CommandLineApplicationExtensions.cs file belongs to neo-express project and is free
+// software distributed under the MIT software license, see the
+// accompanying file LICENSE in the main directory of the
+// repository or https://opensource.org/license/MIT for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using McMaster.Extensions.CommandLineUtils;
+using System.Globalization;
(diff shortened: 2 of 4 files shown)


Wi1l-B0t (approved):
LGTM

ajara87 (approved):

shargon (approved):

PR #740, opened 2026-06-26
Report which batch file line failed

Jim8y:
## Problem

When a command inside a batch file throws (a bad argument, a transfer/deploy failure, a deploy name conflict, etc.), the exception propagates out of the per-line loop up to `OnExecuteAsync`, whose catch calls `app.WriteException(ex)` with **no indication of which line was executing** ([BatchCommand.cs:62-66](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Commands/BatchCommand.cs#L62)). The loop index `i` and the raw line text are never surfaced. For a batch file of dozens of lines, the user sees only the bare exception message and has to guess which command produced it — undercutting the batch feature's whole purpose. The default output gives zero location context (a C# stack trace requires `--stack-trace`).

## Fix

Wrap the per-line parse and dispatch in a `try`/`catch` that rethrows with the **1-based line number and the line text**:

```
Error in batch file line 7: "transfer 10 gas alice bob" - <original message>
```

The original exception is preserved as the inner exception, and `OnExecuteAsync` already prints inner exceptions (`showInnerExceptions: true`), so no detail is lost.

## Note on the diff

Most of the diff is the **indentation change** from wrapping the existing ~190-line `switch` in a `try` block — reviewing with whitespace changes hidden shows the real change is just the added `try {` / `catch { … }`. The switch body is otherwise untouched.

This is on the batch execution path, which has no unit-test harness in the repo (it requires a loaded chain), so it is verified by Release build, `dotnet format --verify-no-changes`, and the existing `BatchCommandParserTests` suite still passing.

### DIFF
--- a/src/neoxp/Commands/BatchCommand.cs
+++ b/src/neoxp/Commands/BatchCommand.cs
@@ -110,230 +110,237 @@ namespace NeoExpress.Commands
                     || args[0].StartsWith("//"))
                     continue;
 
-                var pr = batchApp.Parse(args);
-
-                // Parse() does not run DataAnnotations validation (only ExecuteAsync does),
-                // so enforce the [Required]/[AllowedValues] attributes on the batch models
-                // here. Otherwise a line missing a required argument is dispatched with the
-                // field defaulted to "" and fails later with a confusing downstream error.
-                var validationResult = pr.SelectedCommand.GetValidationResult();
-                if (validationResult != ValidationResult.Success)
-                    throw new Exception(validationResult?.ErrorMessage ?? $"Invalid batch command: {commands.Span[i]}");
-
-                switch (pr.SelectedCommand)
+                try
                 {
-                    case CommandLineApplication<BatchFileCommands.Checkpoint.Create> cmd:
-                        {
-                            _ = await chainManager.CreateCheckpointAsync(
-                                txExec.ExpressNode,
-                                cmd.Model.Name,
-                                cmd.Model.Force,
-                                writer).ConfigureAwait(false);
-                            break;
-                        }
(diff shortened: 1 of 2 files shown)


src/neoxp/Commands/BatchCommand.cs
```diff
@@ -108,195 +108,202 @@ internal async Task ExecuteAsync(IDirectoryInfo root, ReadOnlyMemory<string> com
...
-                            chainManager.SaveChain(chainFilename!);
-                            await writer.WriteLineAsync($"Created Wallet {cmd.Model.Name}");
-                            for (int x = 0; x < wallet.Accounts.Count; x++)
-                                await writer.WriteLineAsync($"    Address: {wallet.Accounts[x].ScriptHash}");
-                            break;
-                        }
-                    default:
-                        throw new Exception($"Unknown batch command {pr.SelectedCommand.GetType()}");
+                        case CommandLineApplication<BatchFileCommands.Contract.Run> cmd:
+                            {
```
Wi1l-B0t, line 195: This `{ }` is necessary?

ajara87:
conflicts

shargon (dismissed):

shargon:
conflicts

src/neoxp/Commands/BatchCommand.cs
Jim8y: Yes, this block scopes the local `script` variable; the preceding `Contract.Invoke` case declares another `script` in the same switch. I left it consistent with the surrounding case bodies.

Jim8y:
Updated.

Wi1l-B0t (approved):

ajara87 (approved):

shargon (approved):

PR #742, opened 2026-06-26
Emit LCOV coverage alongside Cobertura

Jim8y:
## Motivation

The coverage collector ([CodeCoverageDataCollector.cs:117-118](https://github.com/neo-project/neo-express/blob/master/src/collector/CodeCoverageDataCollector.cs#L117)) emits Cobertura XML and a raw text dump. **LCOV** is the other widely-consumed coverage format — genhtml, Coveralls, Codecov, and many editor coverage gutters read it natively. Emitting it directly removes the need for a Cobertura→LCOV conversion step in CI.

## Change

Add `LcovFormat : ICoverageFormat` and wire it into the collector next to the existing formats, producing `neo-coverage.lcov.info`.

The LCOV writer reuses the **same** per-address `HitMap` and sequence-point data the Cobertura writer uses, so the numbers are consistent between the two reports. It groups methods by source document and emits:
- function coverage — `FN`/`FNDA`/`FNF`/`FNH`
- line coverage — `DA` (per source line, highest hit count when several sequence points map to one line), `LF`/`LH`
- `end_of_record` per source file

Branch coverage is intentionally left to the Cobertura report; this first version covers the universally-consumed line and function records.

## Tests

`LcovFormatTests` runs the format against the existing `registrar` coverage fixture (`run1`) and asserts the output is well-formed (`SF:`/`DA:`/`end_of_record`, one `end_of_record` per `SF:`) and that lines-hit never exceeds lines-found per record. Release build is clean, `dotnet format --verify-no-changes` reports no changes, and the full `test-collector` suite passes.

### DIFF
--- a/src/collector/Formats/LcovFormat.cs
+++ b/src/collector/Formats/LcovFormat.cs
@@ -25,12 +25,11 @@ namespace Neo.Collector.Formats
         {
             writeAttachement("neo-coverage.lcov.info", stream =>
             {
-                var writer = new StreamWriter(stream);
+                using var writer = new StreamWriter(stream, leaveOpen: true);
                 foreach (var contract in coverage)
                 {
                     WriteContract(writer, contract);
                 }
-                writer.Flush();
             });
         }
 
@@ -50,7 +49,9 @@ namespace Neo.Collector.Formats
                 writer.WriteLine($"SF:{group.Key}");
 
                 var methods = group.Where(m => m.SequencePoints.Count > 0).ToList();
-                var fnHit = 0;
+                var functionsHit = 0;
+                // LCOV tracefiles use fixed labels: FN=function, FNDA=function data,
+                // FNF=functions found, and FNH=functions hit.
                 foreach (var method in methods)
                 {
                     writer.WriteLine($"FN:{method.SequencePoints[0].Start.Line},{method.Name}");
@@ -59,11 +60,11 @@ namespace Neo.Collector.Formats
                 {
                     var hits = HitsAt(method.SequencePoints[0].Address);
                     if (hits > 0)
-                        fnHit++;
+                        functionsHit++;
                     writer.WriteLine($"FNDA:{hits},{method.Name}");
                 }
                 writer.WriteLine($"FNF:{methods.Count}");
-                writer.WriteLine($"FNH:{fnHit}");
+                writer.WriteLine($"FNH:{functionsHit}");
 
                 // Aggregate per source line; a line can map to several sequence points,
                 // so record the highest hit count observed for that line.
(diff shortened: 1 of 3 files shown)


src/collector/Formats/LcovFormat.cs
```diff
@@ -0,0 +1,106 @@
...
+                    writer.WriteLine($"FN:{method.SequencePoints[0].Start.Line},{method.Name}");
+                }
+                foreach (var method in methods)
+                {
+                    var hits = HitsAt(method.SequencePoints[0].Address);
+                    if (hits > 0)
+                        fnHit++;
+                    writer.WriteLine($"FNDA:{hits},{method.Name}");
+                }
+                writer.WriteLine($"FNF:{methods.Count}");
```
Wi1l-B0t, line 66: What are FNDA, FNF, FNH ?

src/collector/Formats/LcovFormat.cs
```diff
@@ -0,0 +1,106 @@
...
+    // Emits line and function coverage in the LCOV tracefile format consumed by
+    // genhtml, Coveralls, Codecov and many editor coverage gutters. Branch coverage
+    // is intentionally omitted here; it is available in the Cobertura report.
+    class LcovFormat : ICoverageFormat
+    {
+        public void WriteReport(IReadOnlyList<ContractCoverage> coverage, Action<string, Action<Stream>> writeAttachement)
+        {
+            writeAttachement("neo-coverage.lcov.info", stream =>
+            {
+                var writer = new StreamWriter(stream);
```
ajara87:
```suggestion
                using var writer = new StreamWriter(stream);
```

src/collector/Formats/LcovFormat.cs
ajara87: - FNDA is hit counts of a function
- FNF is functions found (total number of functions)
- FNH is functions hit (functions executed at least once)

Wi1l-B0t: Don't use abbreviation.

Jim8y: Addressed: renamed the local counter from `fnHit` to `functionsHit`.

src/collector/Formats/LcovFormat.cs
Jim8y: Addressed with `new StreamWriter(stream, leaveOpen: true)`. I kept `leaveOpen` so disposing the writer does not close the attachment stream owned by `WriteAttachment`.

ajara87 (approved):

Wi1l-B0t (approved):

shargon (approved):

PR #758, opened 2026-06-30
Validate the --additional-gas value before applying it

Jim8y:
## Problem

Both `OnlineNode` and `OfflineNode` apply the `--additional-gas` value as ([OnlineNode.cs:91](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/OnlineNode.cs#L91), [OfflineNode.cs:172](https://github.com/neo-project/neo-express/blob/master/src/neoxp/Node/OfflineNode.cs#L172)):

```csharp
tx.SystemFee += (long)additionalGas.ToBigInteger(NativeContract.GAS.Decimals);
```

Two bad-input cases crash with opaque exceptions:
- **More than 8 decimal places** (e.g. `--gas 0.123456789`) — `ToBigInteger` throws a raw `ArgumentException`.
- **A large value** (e.g. `--gas 100000000000`, any value >= ~92.3e9 GAS) — the scaled `BigInteger` exceeds `Int64.MaxValue` and the `(long)` cast throws `OverflowException`.

Both surface to the user as unhandled, opaque crashes.

## Fix

Route both call sites through a shared `NodeUtility.AdditionalGasSystemFee` helper that returns the `Int64` fee delta, reporting too-many-decimals and too-large values with clear messages (and returning `0` for a non-positive value, as before).

## Tests

Added `AdditionalGasTests`: `1.5` -> `150_000_000`; `0`/`-1` -> `0`; `0.123456789` -> "decimal places" error; `100_000_000_000` -> "too large" error. Release build is clean and format reports no changes.

### DIFF
--- a/src/neoxp/Node/NodeUtility.cs
+++ b/src/neoxp/Node/NodeUtility.cs
@@ -50,11 +50,11 @@ namespace NeoExpress.Node
             }
             catch (ArgumentException)
             {
-                throw new Exception($"--gas value {additionalGas} cannot have more than {NativeContract.GAS.Decimals} decimal places");
+                throw new Exception($"--additional-gas value {additionalGas} cannot have more than {NativeContract.GAS.Decimals} decimal places");
             }
 
             if (value > long.MaxValue)
-                throw new Exception($"--gas value {additionalGas} is too large");
+                throw new Exception($"--additional-gas value {additionalGas} is too large");
 
             return (long)value;
         }
--- a/src/neoxp/Node/NodeUtility.cs
+++ b/src/neoxp/Node/NodeUtility.cs
@@ -34,6 +34,31 @@ namespace NeoExpress.Node
     {
         internal static string ContractNotFoundMessage(UInt160 scriptHash) => $"Contract {scriptHash} not found";
 
+        // Convert an --additional-gas amount to the system-fee delta (in GAS datoshi),
+        // with clear errors instead of the raw exceptions the bare conversion throws: an
+        // ArgumentException for more than GAS.Decimals fractional digits, and an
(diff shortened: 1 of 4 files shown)


shargon (dismissed):

src/neoxp/Node/NodeUtility.cs
```diff
@@ -34,6 +34,31 @@ internal class NodeUtility
...
+                return 0;
+
+            BigInteger value;
+            try
+            {
+                value = additionalGas.ToBigInteger(NativeContract.GAS.Decimals);
+            }
+            catch (ArgumentException)
+            {
+                throw new Exception($"--gas value {additionalGas} cannot have more than {NativeContract.GAS.Decimals} decimal places");
```
Wi1l-B0t:
```suggestion
                throw new Exception($"--additional-gas value {additionalGas} cannot have more than {NativeContract.GAS.Decimals} decimal places");
```
`--additional-gas` ?

src/neoxp/Node/NodeUtility.cs
```diff
@@ -34,6 +34,31 @@ internal class NodeUtility
...
+            }
+
+            if (value > long.MaxValue)
+                throw new Exception($"--gas value {additionalGas} is too large");
```
Wi1l-B0t: Ditto.

Jim8y: Addressed.

src/neoxp/Node/NodeUtility.cs
Jim8y: Addressed.

shargon (approved):

Wi1l-B0t (approved):

ajara87 (approved):

PR #765, opened 2026-07-02
Add a --json option to show balance, show balances and show nft

Jim8y:
## Summary

`show balance`, `show balances` and `show nft` were the only `show` subcommands that emitted prose-only output (`show block`, `show transaction` and `show notifications` already print JSON). A script that submits a transfer with `--json` still had to scrape the resulting balance out of formatted text:

```
GAS (0xd2a4cff31913016155e38e474a2c06d08be276cf)
  balance: 100.5
```

## Change

Add the standard `-j|--json` option to all three subcommands, following the `wallet list` pattern (`JsonTextWriter` over `console.Out`, `Formatting.Indented`):

- `show balance --json` — a single object: `{"symbol", "script-hash", "decimals", "balance"}`
- `show balances --json` — an array of the same shape; an empty array when the account holds nothing
- `show nft --json` — an array of `{"token-id-base64", "token-id-hex"}` objects

Text output is byte-for-byte unchanged. The JSON writers are factored as internal static helpers so their shapes are unit-testable, and `docs/command-reference.md` documents the new option on all three commands.

## Testing

- New `ShowJsonOutputTests` (5 tests) parse the emitted JSON and assert every field, including the empty-array cases.
- Full `test.workflowvalidation` suite: 137 passed.
- Live smoke: `neoxp show balance --help` renders `-j|--json  Output as JSON`.

### DIFF
--- a/src/neoxp/Commands/ShowCommand.Balance.cs
+++ b/src/neoxp/Commands/ShowCommand.Balance.cs
@@ -10,6 +10,8 @@
 
 using McMaster.Extensions.CommandLineUtils;
 using Neo;
+using NeoExpress.Models;
+using Newtonsoft.Json;
 using System.ComponentModel.DataAnnotations;
 
 namespace NeoExpress.Commands
@@ -37,6 +39,9 @@ namespace NeoExpress.Commands
             [Option(Description = "Path to neo-express data file")]
             internal string Input { get; init; } = string.Empty;
 
+            [Option(Description = "Output as JSON")]
+            internal bool Json { get; init; } = false;
+
             internal async Task<int> OnExecuteAsync(CommandLineApplication app, IConsole console)
             {
                 try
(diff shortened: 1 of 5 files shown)


src/neoxp/Commands/ShowCommand.Balance.cs
```diff
@@ -63,6 +76,20 @@ internal async Task<int> OnExecuteAsync(CommandLineApplication app, IConsole con
                     return 1;
                 }
             }
+
+            internal static void WriteBalanceJson(JsonTextWriter writer, Nep17Contract contract, BigDecimal balance)
+            {
+                writer.WriteStartObject();
+                writer.WritePropertyName("symbol");
+                writer.WriteValue(contract.Symbol);
+                writer.WritePropertyName("script-hash");
```
shargon: Use camelcase?

Jim8y: Addressed. The new JSON fields now use camelCase.

shargon (approved):

Wi1l-B0t (approved):

PR #768, opened 2026-07-02
Key the memory storage cache by contract and key, not a 32-bit hash

Jim8y:
## Summary

`MemoryCacheClient` — the default `StateServiceStore` cache used by worknet and checkpoint-based state reads — used `ConcurrentDictionary<int, ...>` where the `int` key was `HashCode.ToHashCode()` over the contract hash and storage key:

```csharp
readonly ConcurrentDictionary<int, byte[]?> storageMap = new();
...
var hash = GetStorageKey(contractHash, key);   // HashCode over (contract, key)
return storageMap.TryGetValue(hash, out value);
```

The 32-bit hash was the cache *identity*, not a bucket. When two distinct `(contract, key)` pairs collide:

- `TryGetCachedStorage` silently returns the other key's cached value (wrong contract storage data), or
- `CacheStorage` throws `Key already exists <hex>` for a key that was never cached.

The birthday bound makes this realistic: the regression test finds a genuine colliding pair within ~100k generated keys. `RocksDbCacheClient` (the persistent variant) already keys by the full contract hash + key bytes, confirming the memory variant was an oversight.

## Change

- `storageMap` is keyed by `(UInt160 contractHash, ReadOnlyMemory<byte> key)` with a comparer combining `UInt160` equality and byte-content equality (`MemorySequenceComparer`). The key bytes are copied on add, so a caller reusing its buffer cannot corrupt the stored identity.
- `foundStateMap` is keyed by the `(UInt160 contractHash, byte? prefix)` pair directly (value-type tuple equality is exact).

The class is `internal`; no public API changes.

## Testing

- `distinct_keys_that_collide_under_a_32bit_hash_keep_their_own_values` reproduces the old hash and finds a real colliding pair, then asserts each key reads back its own value. On the unpatched code it fails with `Key already exists 6B65792D3939303838` — the collision manifesting; with the fix it passes.
- Additional tests: same key under different contracts stays separate, cached identity is content-based (immune to caller buffer mutation), found-states separated per contract/prefix.
- Full `StateServiceStore`/cache filter (7 passed), `dotnet build -c Release`, `dotnet format --verify-no-changes` clean.

### DIFF
--- a/src/bctklib/persistence/StateServiceStore.MemoryCacheClient.cs
+++ b/src/bctklib/persistence/StateServiceStore.MemoryCacheClient.cs
@@ -8,6 +8,7 @@
 // Redistribution and use in source and binary forms with or without
 // modifications are permitted.
 
+using Neo.BlockchainToolkit.Utilities;
 using System.Collections.Concurrent;
 
 namespace Neo.BlockchainToolkit.Persistence
@@ -16,8 +17,11 @@ namespace Neo.BlockchainToolkit.Persistence
     {
         internal sealed class MemoryCacheClient : ICacheClient
         {
-            readonly ConcurrentDictionary<int, byte[]?> storageMap = new();
-            readonly ConcurrentDictionary<int, IList<(ReadOnlyMemory<byte>, byte[])>> foundStateMap = new();
+            // Key the caches by the actual (contract, key) pair rather than by its 32-bit
+            // hash code. Using the hash as the identity meant two distinct storage keys
+            // could collide and silently serve each other's cached value.
+            readonly ConcurrentDictionary<(UInt160 contractHash, ReadOnlyMemory<byte> key), byte[]?> storageMap = new(StorageKeyComparer.Instance);
+            readonly ConcurrentDictionary<(UInt160 contractHash, byte? prefix), IList<(ReadOnlyMemory<byte>, byte[])>> foundStateMap = new();
             private bool disposed;
 
             public void Dispose()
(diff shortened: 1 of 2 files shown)


src/bctklib/persistence/StateServiceStore.MemoryCacheClient.cs
```diff
@@ -16,8 +17,11 @@ public sealed partial class StateServiceStore
     {
         internal sealed class MemoryCacheClient : ICacheClient
         {
-            readonly ConcurrentDictionary<int, byte[]?> storageMap = new();
-            readonly ConcurrentDictionary<int, IList<(ReadOnlyMemory<byte>, byte[])>> foundStateMap = new();
+            // Key the caches by the actual (contract, key) pair rather than by its 32-bit
+            // hash code. Using the hash as the identity meant two distinct storage keys
+            // could collide and silently serve each other's cached value.
+            readonly ConcurrentDictionary<(UInt160 contractHash, ReadOnlyMemory<byte> key), byte[]?> storageMap = new(StorageKeyComparer.Instance);
+            readonly ConcurrentDictionary<(UInt160 contractHash, byte? prefix), IList<(ReadOnlyMemory<byte>, byte[])>> foundStateMap = new();
```
shargon, line 24: The dictionary doesn't have a comparer?

Jim8y: The storage map uses StorageKeyComparer. This map is only keyed by contract hash and the one-byte prefix, so the default tuple comparer is enough.

shargon (approved):

Wi1l-B0t (approved):

PR #769, opened 2026-07-02
Add a neo-worknet stop command

Jim8y:
## Summary

Neo-WorkNet's RPC server has always exposed the `expressshutdown` method (`WorknetRpcServerPlugin.ExpressShutdown`), but no CLI command used it — the only way to stop a running worknet node was CTRL-C in its own terminal. Neo-Express has had `neoxp stop` for the equivalent workflow from the start.

## Change

Add `neo-worknet stop`, mirroring `ExpressChainManager.StopNodeAsync` from Neo-Express:

- loads the worknet file (`--input` arrives via the existing `InputFileConvention`),
- sends `expressshutdown` to the node's RPC endpoint (`localhost:30332`, the port `run` hardcodes),
- waits for the returned `process-id` to exit, then reports `worknet node stopped`;
- when no node is listening, reports `worknet node was not running` and exits 0 instead of failing.

Documented in `docs/worknet-command-reference.md`.

## Testing

Neo-WorkNet has no test project, so verified end-to-end against the built binary:

- `neo-worknet --help` lists `stop`; `stop --help` shows the auto-added `--input` option.
- Negative path: `stop` against a worknet file with no node running prints `worknet node was not running`, exit 0.
- Positive path: against a stub JSON-RPC server on 30332 answering `expressshutdown` with the pid of a short-lived process, `stop` waited for that process to exit (wall clock matched the process lifetime) and printed `worknet node stopped`, exit 0.


shargon (dismissed):

src/worknet/Commands/StopCommand.cs
```diff
@@ -0,0 +1,68 @@
...
+            using var rpcClient = new RpcClient(new Uri($"http://localhost:{RPC_PORT}"),
+                protocolSettings: worknet.BranchInfo.ProtocolSettings);
+            Neo.Json.JToken json;
+            try
+            {
+                json = await rpcClient.RpcSendAsync("expressshutdown").ConfigureAwait(false);
+            }
+            catch (HttpRequestException)
+            {
+                await console.Out.WriteLineAsync("worknet node was not running").ConfigureAwait(false);
```
Wi1l-B0t: No exception message ?

ajara87:
@Jim8y conflicts

src/worknet/Commands/StopCommand.cs
Jim8y: Updated.

Jim8y:
Updated, conflicts are resolved.

src/worknet/Commands/StopCommand.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                json = await rpcClient.RpcSendAsync("expressshutdown").ConfigureAwait(false);
+            }
+            catch (HttpRequestException ex)
+            {
+                await console.Out.WriteLineAsync($"worknet node was not running: {ex.Message}").ConfigureAwait(false);
+                return 0;
+            }
+
+            var processId = int.Parse(json["process-id"]!.AsString());
```
shargon: seems that we are using different case in different areas

shargon (dismissed):

src/worknet/Commands/StopCommand.cs
Jim8y: Updated.

shargon (dismissed):

src/worknet/Commands/StopCommand.cs
shargon: I mean the json "process-id" > "processId"

ajara87:
UT Fails

src/worknet/Commands/StopCommand.cs
Jim8y: Updated.

Jim8y:
Fixed; CI is green now.

ajara87 (approved):

shargon (approved):

PR #772, opened 2026-07-02
Share a single SimpleJSON source between collector and build-tasks

Jim8y:
## Summary

`src/collector/Utility/SimpleJSON.cs` and `src/build-tasks/SimpleJSON.cs` were byte-identical 1450-line copies of the vendored SimpleJSON parser (`diff -q` confirms). Two copies invite silent divergence: any fix applied to one must be remembered and mirrored by hand in the other.

Neither project can take the dependency another way — `Neo.BuildTasks` is a standalone MSBuild task package (net10.0 + netstandard2.0, no ProjectReferences allowed in its packaging model) and `Neo.Collector` is a VSTest data collector whose only dependency is `Microsoft.TestPlatform.ObjectModel` — so a shared *source* file is the right mechanism.

## Change

- `git mv src/build-tasks/SimpleJSON.cs src/shared/SimpleJSON.cs`, delete the collector copy.
- Both csprojs compile the shared file via `<Compile Include="..\shared\SimpleJSON.cs" Link="..." />` with a comment noting it is shared.

The declared namespace (`SimpleJSON`) and API are untouched, so no consuming code changes; both projects compile exactly the bytes they compiled before.

## Testing

- `dotnet build` of both projects (all target frameworks, including build-tasks' netstandard2.0) and the full solution: clean.
- `test-build-tasks` (35 passed) and `test-collector` (11 passed) unchanged.
- `grep` confirms no other references to the old paths remain.


shargon (approved):
It's not better a library?

Jim8y:
For now I kept it as shared source because both projects need the same internal parser without adding a public package dependency. If it grows, we can split it into a small library later.

Wi1l-B0t (approved):

PR #777, opened 2026-07-02
Add CheckpointFixture.CreateEngine to cut contract-test setup ceremony

Jim8y:
## Summary

Every harness-based contract test repeats the same preamble (both sample tests in `samples/test/contract-tests.cs` duplicate it verbatim):

```csharp
var settings = chain.GetProtocolSettings();
using var snapshot = fixture.GetSnapshot();
using var engine = new TestApplicationEngine(snapshot, settings);
```

— even resolving settings the roundabout way through `FindChain().GetProtocolSettings()` when `CheckpointFixture.ProtocolSettings` already exposes the checkpoint's own settings. A Foundry/Anchor test body starts in one call.

## Change

Add a `CreateEngine` extension on `CheckpointFixture`:

```csharp
using var engine = fixture.CreateEngine();                       // default test transaction
using var engine = fixture.CreateEngine(signer, WitnessScope.Global);  // signed variant
```

It snapshots the checkpoint (the existing `GetSnapshot` path), builds a `TestApplicationEngine` with the fixture's own `ProtocolSettings`, and returns an engine that owns the snapshot — its `Dispose` also disposes the snapshot, so one `using` line replaces the three-line ceremony. The explicit `GetSnapshot` path is unchanged.

The samples still build against published packages, so they are left as-is; they can adopt the one-line form once a package with this API ships.

## Testing

New `CreateEngineTests` build a real checkpoint at runtime (RocksDB → genesis-initialized ledger → `RocksDbUtility.CreateCheckpoint`) and drive the full path through a `CheckpointFixture`:

- the engine carries the checkpoint's network/address-version and executes a script to `HALT`;
- the signer overload produces a transaction container with the requested account and witness scope;
- engines are independent, and double-dispose is safe.

Full `test-harness` suite (10 passed), `dotnet build -c Release` and `dotnet format --verify-no-changes` clean.

### DIFF
--- a/src/test-harness/Extensions.cs
+++ b/src/test-harness/Extensions.cs
@@ -11,8 +11,10 @@
 using Neo;
 using Neo.BlockchainToolkit;
 using Neo.BlockchainToolkit.Persistence;
+using Neo.BlockchainToolkit.SmartContract;
 using Neo.BlockchainToolkit.Utilities;
 using Neo.Extensions;
+using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
 using Neo.SmartContract;
 using Neo.SmartContract.Native;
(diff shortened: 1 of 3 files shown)


src/test-harness/Extensions.cs
```diff
@@ -248,5 +250,37 @@ public static StoreCache GetSnapshot(this CheckpointFixture fixture)
...
+        public static TestApplicationEngine CreateEngine(this CheckpointFixture fixture, UInt160? signer = null, WitnessScope witnessScope = WitnessScope.CalledByEntry)
+        {
+            var snapshot = fixture.GetSnapshot();
+            var container = signer is null
+                ? TestApplicationEngine.CreateTestTransaction()
+                : TestApplicationEngine.CreateTestTransaction(signer, witnessScope);
+            return new FixtureApplicationEngine(snapshot, fixture.ProtocolSettings, container);
+        }
+
+        sealed class FixtureApplicationEngine : TestApplicationEngine
```
shargon: I think applicationEngine already have disposables inside, maybe we can extend the ApplicationEngine for add entries there

Jim8y: Addressed. TestApplicationEngine can now own extra disposables, and CreateEngine registers the snapshot there instead of using a subclass.

shargon (approved):

Wi1l-B0t (approved):

PR #780, opened 2026-07-02
Merge tracking-store seek results lazily instead of sorting them

Jim8y:
## Summary

The three tracking-store `Seek` implementations (`PersistentTrackingStore`, its `Snapshot`, and `MemoryTrackingStore`, which also serves its snapshot) combined tracked and backing-store items with:

```csharp
return trackedItems.Concat(storeItems).OrderBy(kvp => kvp.Key, comparer);
```

`OrderBy` buffers the entire concatenated sequence on its first pull. A prefixed `DataCache.Find` that reads a handful of entries — the dominant access pattern during contract execution — therefore enumerated the backing store **to its end** and paid an O(n log n) sort on every call. For `PersistentTrackingStore` each store item also costs a RocksDB point lookup (`KeyUntracked`), so the waste multiplies. These stores back the neoxp offline node, worknet's run/fastfwd stores, and the checkpoint-based test harness.

## Change

Both inputs are already sorted in seek direction — the RocksDB tracking iterator and the backing store's `Find` contract — so replace the sort with a lazy two-way merge (`SortedMerge.MergeSorted`) that enumerates each side only as far as the caller consumes. The dictionary-backed sides (the in-memory tracking map, a snapshot's uncommitted changes) are unsorted, so those small sides are ordered individually before merging with the already-sorted store side.

Semantics are unchanged: the callers filter duplicate keys between the two sides before merging (deletion masking and tracked-wins behavior live in those filters, not in the sort), and the merge preserves first-side preference on ties regardless.

## Verification

- **Laziness proven by test**: `tracking_store_find_no_longer_walks_the_backing_store_to_its_end` puts 1,000 entries in an instrumented backing store and takes 5 items from a tracking-store `Find`. On the old code it fails with `found 1000 (difference of 980)` items pulled; with the merge it reads fewer than 20.
- **Behavior preserved**: the full `test.bctklib` suite — 356 tests covering forward/backward seeks, prefix seeks, deletion masking, update shadowing and snapshot isolation across Memory/Persistent/Checkpoint store variants — passes unchanged.
- New unit tests cover forward/backward merged ordering, tie preference, and merge laziness directly.

### DIFF
--- a/src/bctklib/persistence/PersistentTrackingStore.cs
+++ b/src/bctklib/persistence/PersistentTrackingStore.cs
@@ -166,7 +166,9 @@ namespace Neo.BlockchainToolkit.Persistence
                 ? MemorySequenceComparer.Default
                 : MemorySequenceComparer.Reverse;
 
-            return trackedItems.Concat(storeItems).OrderBy(kvp => kvp.Key, comparer);
+            // both sides are already sorted in seek direction (RocksDB iterator and the
+            // backing store's Find); merge lazily instead of buffering and re-sorting
+            return trackedItems.MergeSorted(storeItems, comparer);
 
             static IEnumerable<(byte[] Key, byte[] Value)> SeekTracked(
                 byte[]? key, SeekDirection direction, RocksDb db,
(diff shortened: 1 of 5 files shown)


src/bctklib/persistence/PersistentTrackingStore.cs
```diff
@@ -166,7 +166,9 @@ static bool Contains(byte[]? key, RocksDb db, ColumnFamilyHandle columnFamily, R
                 ? MemorySequenceComparer.Default
                 : MemorySequenceComparer.Reverse;
 
-            return trackedItems.Concat(storeItems).OrderBy(kvp => kvp.Key, comparer);
+            // both sides are already sorted in seek direction (RocksDB iterator and the
+            // backing store's Find); merge lazily instead of buffering and re-sorting
+            return trackedItems.MergeSorted(storeItems, comparer);
```
shargon, line 174: Here storedItems are not ordered before, like in the other change

Jim8y: Updated to order the backing-store side before merging, so the result keeps the same ordering guarantees as the previous Concat+OrderBy path. Related tracking-store tests pass.

shargon (approved):

ajara87 (approved):

PR #781, opened 2026-07-02
Resolve keyed state reads with a point lookup instead of a scan

Jim8y:
## Summary

`StateServiceStore` resolved a single storage key for deployed contracts by enumerating the contract's **entire** cached record set:

```csharp
byte[]? GetFromStates(UInt160 contractHash, byte? prefix, ReadOnlyMemory<byte> key)
{
    return FindStates(contractHash, prefix)
        .FirstOrDefault(kvp => MemorySequenceComparer.Equals(kvp.key.Span, key.Span)).value;
}
```

Every `ApplicationEngine` storage read during worknet execution pays this, scaling with the contract's record count. For `RocksDbCacheClient` (the worknet `run`/`fastfwd` path) the enumeration walks the whole column family allocating a key and a value array per record visited — even though that column family is keyed by the exact storage key and the same file already uses direct `db.GetSlice` point lookups for the singleton-storage cache.

## Change

- `ICacheClient.TryGetCachedState(contractHash, prefix, key, out value)` — returns `false` when the record set has not been cached yet, `true` with `null` for a definitive miss.
- `RocksDbCacheClient` answers with a single `GetSlice` on the found-states column family.
- `MemoryCacheClient` indexes each record set by key content (`MemorySequenceComparer`) when the download snapshot commits; the download-ordered list is kept for the enumerating API.
- `GetFromStates` now: point lookup → on first use, download the record set (existing `DownloadStates` path, unchanged) → point lookup. `Seek`/`Find` still use the enumerating API.

## Verification

- New `CacheClientPointLookupTests` (5 tests): on **both** cache clients, the point lookup returns exactly what the enumerating API returns for every seeded record; a cached-but-absent key is a definitive `null`; an undownloaded record set reports unanswerable (`false`) rather than an empty hit; prefixed and unprefixed record sets stay separate.
- Full `test.bctklib` suite: 356 passed unchanged (covers the `StateServiceStore` read paths end to end).

Note: touches the same `MemoryCacheClient` file as #768; whichever merges second I will rebase.

### DIFF
--- a/src/bctklib/persistence/StateServiceStore.MemoryCacheClient.cs
+++ b/src/bctklib/persistence/StateServiceStore.MemoryCacheClient.cs
@@ -8,6 +8,7 @@
 // Redistribution and use in source and binary forms with or without
 // modifications are permitted.
 
+using Neo.BlockchainToolkit.Utilities;
 using System.Collections.Concurrent;
 
 namespace Neo.BlockchainToolkit.Persistence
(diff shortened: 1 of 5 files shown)


src/bctklib/persistence/StateServiceStore.MemoryCacheClient.cs
```diff
@@ -17,9 +18,28 @@ public sealed partial class StateServiceStore
...
+            // instead of a scan.
+            sealed class FoundStates
+            {
+                public readonly IReadOnlyList<(ReadOnlyMemory<byte> key, byte[] value)> Items;
+                public readonly Dictionary<ReadOnlyMemory<byte>, byte[]> Index;
+
+                public FoundStates(List<(ReadOnlyMemory<byte> key, byte[] value)> items)
+                {
+                    Items = items;
+                    Index = new Dictionary<ReadOnlyMemory<byte>, byte[]>(items.Count, MemorySequenceComparer.Default);
```
Wi1l-B0t, line 35:
```suggestion
                    Index = new(items.Count, MemorySequenceComparer.Default);
```

ajara87:
conflicts

Jim8y:
This is now covered by master. Exact keyed reads use GetProvenState directly, so keeping this PR would reintroduce findstates for exact reads. Closing it.

PR #782, opened 2026-07-02
Seed the worknet consensus account with GAS at branch time

Jim8y:
## Summary

After branching a public network, the only account the user controls — the generated consensus account — holds **zero GAS**, so the fork is read-only in practice: nothing can be deployed or invoked until committee emission slowly accrues one block at a time. `CreateCommand.InitializeStore` already rewrites the NEO committee and candidate storage at branch time, but never touched balances. (Anvil and Hardhat both start forks with funded accounts for exactly this reason.)

## Change

- `InitializeStore` seeds each consensus account's GAS balance the same way the committee is patched: an `AccountState` written into GAS native storage (`Prefix_Account`), with `TotalSupply` bumped by the seeded amount so the ledger stays internally consistent. Existing balances are added to, not overwritten.
- Default seed: **10,000 GAS**, configurable with a new `--gas` option on both `create` and `reset` (`--gas 0` disables seeding; negative values are rejected). `reset` re-seeds through the same code path it already used.
- Documented in `docs/worknet-command-reference.md`, including a "Consensus account funding" note.

## Testing

This PR also introduces the **`test.worknet`** test project — the worknet tool previously had no tests at all. The new tests run `InitializeStore` against a genesis-initialized store and read results back through the real native-contract APIs:

- the default seed lands in `GAS.BalanceOf` (10,000 × 10⁸) and `GAS.TotalSupply` grows by exactly that amount;
- a custom `--gas 12.5` seed produces the exact datoshi balance;
- `--gas 0` changes neither balance nor supply;
- the existing behaviors are regression-guarded: the committee is still replaced with the consensus key and the branch block is still appended.

Removing the seeding code makes the suite fail (verified by mutation). `dotnet build` of the full solution (including the new project) and `dotnet format --verify-no-changes` are clean; `create --help`/`reset --help` render the new option.


src/worknet/Commands/CreateCommand.cs
```diff
@@ -163,6 +177,34 @@ internal static void InitializeStore(IStore store, IEnumerable<WalletAccount> co
             snapshot.Delete(key);
         }
 
+        // Seed the consensus accounts with GAS so the branch can pay transaction fees
+        // immediately (contract deploys and invokes) instead of being read-only until
+        // committee emission slowly accrues. Total supply is bumped to match.
+        if (gasSeed > 0)
+        {
```
shargon, line 187: It's not better to do a transfer? This doesn't emit the event

Jim8y: Updated to seed GAS through the native GAS transfer path from the current committee account to the worknet consensus account. This keeps total supply unchanged and uses the normal Transfer logic. test.worknet passes.

docs/worknet-command-reference.md
```diff
@@ -56,6 +57,13 @@ the branched chain. Based on understanding of Neo-Express usage patterns, multip
 not typically used. If four or seven conesnsus node support in Neo-WorkNet is important to you, please
 file an issue in our [GitHub repo](https://github.com/neo-project/neo-express/issues)
 
+### Consensus account funding
+
+When a worknet is created (and whenever it is reset), the generated consensus account is seeded
+with GAS — 10,000 by default, configurable with the `--gas` option — so contracts can be deployed
```
Wi1l-B0t: Which unit?

src/worknet/Commands/CreateCommand.cs
```diff
@@ -70,6 +75,9 @@ internal async Task<int> OnExecuteAsync(CommandLineApplication app, IConsole con
                 DiagnosticListener.AllListeners.Subscribe(diagnosticObserver);
             }
 
+            if (Gas < 0)
+                throw new ArgumentException("--gas cannot be negative");
```
Wi1l-B0t:
```suggestion
            ArgumentOutOfRangeException.ThrowIfNegative(Gas, "--gas cannot be negative");
```

ajara87:
conflicts

docs/worknet-command-reference.md
Jim8y: Updated the docs to state GAS explicitly.

src/worknet/Commands/CreateCommand.cs
Jim8y: Updated.

Jim8y:
Updated, conflicts are resolved.

src/worknet/Commands/ResetCommand.cs
```diff
@@ -27,12 +27,17 @@ public ResetCommand(IFileSystem fs)
...
+    internal decimal Gas { get; init; } = CreateCommand.DEFAULT_GAS_SEED;
+
     internal async Task<int> OnExecuteAsync(CommandLineApplication app, IConsole console, CancellationToken token)
     {
         try
         {
             if (!Force)
                 throw new InvalidOperationException("--force must be specified when resetting worknet");
+            if (Gas < 0)
+                throw new ArgumentException("--gas cannot be negative");
```
ajara87:
```suggestion
            ArgumentOutOfRangeException.ThrowIfNegative(Gas, nameof(Gas));
```

Jim8y: Updated.

src/worknet/Commands/CreateCommand.cs
```diff
@@ -163,6 +179,18 @@ internal static void InitializeStore(IStore store, IEnumerable<WalletAccount> co
             snapshot.Delete(key);
         }
 
+        // Seed the consensus accounts with GAS so the branch can pay transaction fees
+        // immediately (contract deploys and invokes) instead of being read-only until
+        // committee emission slowly accrues. Total supply is bumped to match.
+        if (gasSeed > 0)
+        {
+            var seedAmount = (BigInteger)(gasSeed * 100_000_000m); // GAS has 8 decimals
```
shargon, line 188: Is sent the expected * consensusAccounts

Jim8y: Updated this to make it explicit: gasSeed is applied to each consensus account, and the GAS comes from the current committee account, so total supply does not change. I also added a regression test with two consensus accounts.

test/test.worknet/test.worknet.csproj
```diff
@@ -0,0 +1,26 @@
...
+        <PackageReference Include="FluentAssertions" Version="7.2.0" />
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.0.0" />
+        <PackageReference Include="xunit.runner.visualstudio" Version="3.1.5">
+          <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+          <PrivateAssets>all</PrivateAssets>
+        </PackageReference>
+        <PackageReference Include="xunit.v3" Version="3.1.0" />
+    </ItemGroup>
+    <ItemGroup>
+      <PackageReference Update="Nerdbank.GitVersioning" Version="3.8.118" />
```
shargon: Not needed

Jim8y: Removed.

shargon (approved):

Wi1l-B0t (approved):

PR #783, opened 2026-07-02
Confirm transactions instantly on single-node chains with chain.AutoMine

Jim8y:
## Summary

Against a running node, every transaction — transfers, contract deploys/invokes, `execute`, policy and candidate operations — enters the mempool and waits out the dBFT block interval (15s default, 1s minimum). The offline node already gives instant finality (`OfflineNode.SubmitTransactionAsync` wraps each transaction in a consensus-signed block), so the same command against a *stopped* chain confirms immediately while the dev loop against `neoxp run` pays a full block interval per step. This is the latency Anvil/Hardhat eliminate with automine.

## Change

Add an opt-in **`chain.AutoMine`** setting. On single-node chains — where the express process holds every consensus key — `OnlineNode.ExecuteAsync` wraps the signed transaction in a consensus-signed block (`NodeUtility.CreateSignedBlock`, the code `fastfwd` and the offline node already rely on) and submits the block via `submitblock`, instead of `sendrawtransaction`:

- **Opt-in**: defaults to off; nothing changes without the setting.
- **Single-node only**: ignored on 4/7-node chains (other validators would have to sign).
- **Race-safe**: if the submitted block loses a race with an in-flight dBFT proposal at the same height, it retries once on the new tip, then falls back to normal mempool submission.
- The regular dBFT timer keeps minting empty blocks on its normal schedule alongside.

Documented in `docs/settings.md`.

## Verification

Live A/B test against a running node with a **60-second** block interval:

| | height before → 1s after transfer |
|---|---|
| `"chain.AutoMine": "true"` | 2 → **3** (transaction confirmed instantly, balance visible immediately) |
| no setting (control) | 2 → 2 (mempool behavior unchanged) |

- New `AutoMineTests` (7 cases) cover the gate: off by default, on for single-node + setting, ignored on multi-node chains, ignored for `false`/`0`/invalid/empty values.
- Full non-integration suite: 133 passed. `dotnet build -c Release` and `dotnet format --verify-no-changes` clean.

### DIFF
--- a/src/neoxp/Node/OnlineNode.cs
+++ b/src/neoxp/Node/OnlineNode.cs
@@ -113,6 +113,50 @@ namespace NeoExpress.Node
 
             var tx = await tm.SignAsync().ConfigureAwait(false);
 
+            if (ShouldAutoMine(chain))
+            {
+                return await SubmitAutoMinedBlockAsync(tx).ConfigureAwait(false);
+            }
+
+            return await rpcClient.SendRawTransactionAsync(tx).ConfigureAwait(false);
+        }
+
+        // Instant transaction confirmation is opt-in via the chain.AutoMine setting and only
+        // supported on single-node chains, where the express process holds every consensus key.
+        internal static bool ShouldAutoMine(ExpressChain chain)
+            => chain.ConsensusNodes.Count == 1
+                && chain.TryReadSetting<bool>("chain.AutoMine", bool.TryParse, out var autoMine)
+                && autoMine;
+
+        // Confirm the transaction immediately by wrapping it in a consensus-signed block and
+        // submitting the block, the same way OfflineNode.SubmitTransactionAsync and
+        // FastForwardAsync already do, instead of waiting out the dBFT block interval.
+        // Submitting can race an in-flight dBFT proposal at the same height (the first block
+        // at a height wins), so retry once on the new tip and fall back to the mempool if the
+        // race is lost twice.
+        async Task<UInt256> SubmitAutoMinedBlockAsync(Transaction tx)
+        {
+            for (var attempt = 0; attempt < 2; attempt++)
+            {
+                var prevHash = await rpcClient.GetBestBlockHashAsync().ConfigureAwait(false);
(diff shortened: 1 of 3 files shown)


src/neoxp/Node/OnlineNode.cs
```diff
@@ -113,6 +113,50 @@ public async Task<UInt256> ExecuteAsync(Wallet wallet, UInt160 accountHash, Witn
...
+            }
+
+            return await rpcClient.SendRawTransactionAsync(tx).ConfigureAwait(false);
+        }
+
+        // Instant transaction confirmation is opt-in via the chain.AutoMine setting and only
+        // supported on single-node chains, where the express process holds every consensus key.
+        internal static bool ShouldAutoMine(ExpressChain chain)
+            => chain.ConsensusNodes.Count == 1
+                && chain.TryReadSetting<bool>("chain.AutoMine", bool.TryParse, out var autoMine)
```
shargon, line 130: The config should be cached, is executed in every tx

Jim8y: Updated. The chain.AutoMine setting is now evaluated once when OnlineNode is created and reused for each transaction. Related tests pass.

src/neoxp/Node/OnlineNode.cs
```diff
@@ -113,6 +115,50 @@ public async Task<UInt256> ExecuteAsync(Wallet wallet, UInt160 accountHash, Witn
...
+                    ProtocolSettings.Network,
+                    new[] { tx });
+                try
+                {
+                    await rpcClient.SubmitBlockAsync(block.ToArray()).ConfigureAwait(false);
+                    return tx.Hash;
+                }
+                catch (RpcException)
+                {
+                    // lost the race with a dBFT-proposed block at this height
```
Wi1l-B0t, line 158: Just ignore? No log?

ajara87 (approved):

src/neoxp/Node/OnlineNode.cs
Jim8y: This fallback is intentionally quiet. It only means another block won the same-height race, and the transaction is still submitted through the normal mempool path after retries. Logging every expected race here would be noisy.

shargon (approved):

PR #791, opened 2026-07-02
Fix numeric block lookup in show block

Jim8y:
## Summary
- Parse numeric block identifiers before hash identifiers in show block lookup.
- Add workflow validation coverage for latest, numeric index, and hash lookups.

## Root cause
`show block 0` passed `"0"` into the hash-oriented path before trying block index lookup, so block height 0 could not be resolved reliably.

## Testing
- `dotnet test test/test.workflowvalidation/test.workflowvalidation.csproj --configuration Release --no-restore --filter 'FullyQualifiedName~ExpressNodeExtensionsTests' --verbosity minimal`
- `neoxp create -f`
- `neoxp show block 0`

### DIFF
--- a/src/neoxp/Extensions/ExpressNodeExtensions.cs
+++ b/src/neoxp/Extensions/ExpressNodeExtensions.cs
@@ -489,14 +489,14 @@ namespace NeoExpress
                 return await expressNode.GetLatestBlockAsync().ConfigureAwait(false);
             }
 
-            if (UInt256.TryParse(blockHash, out var uint256))
-            {
-                return await expressNode.GetBlockAsync(uint256).ConfigureAwait(false);
-            }
-
             if (uint.TryParse(blockHash, out var index))
             {
                 return await expressNode.GetBlockAsync(index).ConfigureAwait(false);
+            }
+
+            if (UInt256.TryParse(blockHash, out var uint256))
+            {
+                return await expressNode.GetBlockAsync(uint256).ConfigureAwait(false);
             }
 
             throw new ArgumentException($"{nameof(blockHash)} must be block index, block hash or empty", nameof(blockHash));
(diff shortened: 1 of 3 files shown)


src/neoxp/Extensions/ExpressNodeExtensions.cs
```diff
@@ -489,14 +489,14 @@ public static async Task<Block> GetBlockAsync(this IExpressNode expressNode, str
                 return await expressNode.GetLatestBlockAsync().ConfigureAwait(false);
             }
 
-            if (UInt256.TryParse(blockHash, out var uint256))
+            if (uint.TryParse(blockHash, out var index))
```
Wi1l-B0t, line 492: What is difference between `UInt256 then uint` and `uint then UInt256`?

superboyiii: Seems no difference, it's more like a defensive modification.

Jim8y: The difference is numeric strings like "0". If UInt256 is tried first, "0" is parsed as a hash value and block index 0 cannot be shown. Trying uint first preserves index lookup for numeric inputs.

superboyiii: > The difference is numeric strings like "0". If UInt256 is tried first, "0" is parsed as a hash value and block index 0 cannot be shown. Trying uint first preserves index lookup for numeric inputs.

`UInt256.TryParse` requires exactly 64 hexadecimal characters, so that means a uint with 64 hexadecimal characters and begin with "0", it's impossible for block height. But it's OK to do this check as a defensive modification.

NGDAdmin (approved):

src/neoxp/Extensions/ExpressNodeExtensions.cs
Jim8y: You are right about UInt256.TryParse. The main issue is the command call site: BlockHash is a string, but expressNode.GetBlockAsync(BlockHash) binds to the UInt256 overload through the implicit string -> UInt256 conversion instead of the string extension method. The explicit extension call keeps numeric input on the index/hash parser path, and the uint-first order makes that parser behavior explicit.

ajara87 (approved):

shargon (approved):

PR #793, opened 2026-07-02
Sync runtime reference docs

Jim8y:
## Summary
- Document the supported invocation file contract format.
- Clarify NeoTrace and local trace usage.
- Update runtime/version wording in command and installation docs.

## Root cause
The invocation file docs described an object form with hash/path fields, but the current command resolves `contract` as a string containing a deployed contract name or script hash. A `.nef` path is not enough to identify an already deployed N3 contract hash.

## Testing
- `rg -n "hash/path|path.*nef|contract: \\{|DebugInfo|ConvertScriptToInstructions|wither|3\\.6|coveralls" docs readme.md`
- `git diff --check origin/master..HEAD`


Wi1l-B0t (approved):

NGDAdmin (approved):

shargon (commented):
LGTM

PR #795, opened 2026-07-03
Use two-parameter contract update without data

Jim8y:
Summary:
- treat omitted contract update data as null so two-parameter update methods are used
- apply the same behavior to batch contract update commands
- add unit coverage for update data parsing

Tests:
- dotnet test test/test.workflowvalidation/test.workflowvalidation.csproj --configuration Release --filter "FullyQualifiedName~ContractUpdateCommandTests|FullyQualifiedName~BatchCommandParserTests" --verbosity minimal
- dotnet format neo-express.sln --verify-no-changes --no-restore --verbosity minimal
- manual deploy/update/run validation with a two-parameter update contract
- manual batch update validation without --data

### DIFF
--- a/src/neoxp/Commands/BatchCommand.cs
+++ b/src/neoxp/Commands/BatchCommand.cs
@@ -214,7 +214,7 @@ namespace NeoExpress.Commands
                             }
                         case CommandLineApplication<BatchFileCommands.Contract.Update> cmd:
                             {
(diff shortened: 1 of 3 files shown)


Wi1l-B0t (approved):

NGDAdmin (approved):

shargon (commented):
LGTM

PR #801, opened 2026-07-03
Fix Visual Tracker create output argument

Jim8y:
## Summary
- pass `-o` when Visual Tracker creates a Neo Express instance
- update generated contract tasks/templates to create and update wallets against the intended `.neo-express` file
- add a regression test for the create command arguments

## Validation
- `npm test`
- `npm run compile`

This was found while validating the Visual Tracker Quick Start flow with Neo Express 3.10.0: the old positional output path is rejected by `neoxp create`, which now requires `-o|--output`.


NGDAdmin (approved):

ajara87 (approved):

shargon (commented):
LGTM

PR #803, opened 2026-07-03
Fix Visual Tracker checkpoint restore while running

Jim8y:
## Summary
- stop the selected Neo Express instance before Visual Tracker reset and checkpoint restore operations
- restart the instance after restore/reset when it was running before the operation
- add a regression check for the offline checkpoint flow

## Validation
- npm test
- npm run compile
- npm run typecheck
- manually verified in VS Code with the packaged Visual Tracker extension: create a single-node Neo Express instance, create a checkpoint, fast-forward 3 blocks, restore the checkpoint while the node is running, and confirm the RPC block count returns from 4 to 1 with Neo Express restarted

Targets v3.10.0 validation.


ajara87 (approved):

NGDAdmin (approved):

shargon (commented):
LGTM

PR #835, opened 2026-08-25
Improve VS Code DevTracker, official C# samples, and Neo Express deploy UX

cschuchardt88:
## Summary

This updates Visual DevTracker, the C# samples, and `neoxp` so a local Neo Express workspace can be created, built, deployed, and invoked from VS Code without the UX bugs we hit on current 3.10 / .NET 10.

### Samples and wizard
- Official [Neo.SmartContract.Template](https://github.com/neo-project/neo-devpack-dotnet/tree/master-n3/src/Neo.SmartContract.Template) starters (Blank, NEP-17, NEP-11, Oracle, Ownable) live under `samples/examples/` with `Neo.BuildTasks`.
- `dotnet build` in an example folder creates `default.neo-express` if missing, compiles the `.nef`, and deploys as `genesis` while Express is running.
- The **New contract** wizard overlays those same official C# starters (plus the existing Storage scaffold). Official unit-test projects are not copied.

### Visual DevTracker
- Quick Start is a tree view (sidebar scrollbar, no nested webview bar). **Deploy a contract** stays; **Redeploy** is removed.
- Contract Studio: contract/method dropdowns, scroll on short panes, persistent invocation results, Output channel **Neo Express Invocation**.
- Wallets: genesis + `node1` + NEP-6 accounts. Deploy as genesis uses `--password` when needed.
- Reset stops leftover node processes (including after F5 reload), does not hang on failed start, and clears the status bar.
- F5 uses a repo-built `src/neoxp/bin/.../neoxp.dll` when present, instead of an older global `neoxp` that lacks `--gas`.

### `neoxp` / BuildTasks
- Deploy system fee is padded (invokescript underestimates persist-time `_deploy` cost by a few hundred datoshi → `Insufficient GAS` FAULT). Default `--gas 1` plus a 0.1 GAS estimate pad.
- After deploy/update, wait for the application log; FAULT is an error, not “Transaction submitted”.
- `deploy --force` on the **same hash** calls the contract `update` method (which must call `ContractManagement.Update`). No `update` method → error telling you to reset or restore a checkpoint.
- Clean/rebuild deletes the batch stamp so the next build resets and redeploys.

### Docs
- New [docs/getting-started.md](docs/getting-started.md) (CLI and VS Code).
- Debugger command reference and sample README updates.

### Prerequisites
- [.NET 10 SDK](https://dotnet.microsoft.com/download/dotnet/10.0) (`dotnet --list-sdks` lists a 10.x).
- Node.js 20+ (for the extension).
[...]

PR #836, opened 2026-08-25
fix(neoxp): pad deploy fees, wait for FAULT, and honor force-update (review 1st)

cschuchardt88:
## Summary

CLI/BuildTasks half of what was #835.

`neoxp contract deploy` used RPC `invokescript` `GasConsumed` as `SystemFee`. Persist-time `_deploy` is a few hundred datoshi higher, so the node FAULTed with `Insufficient GAS` while the CLI printed “Transaction submitted”.

### Changes
- Pad system fee by 0.1 GAS plus `--gas` (deploy defaults to 1).
- After deploy/update, wait for the application log; **FAULT is an error**.
- `deploy --force` on the **same hash** calls the contract `update` method (`ContractManagement.Update` from inside the contract). No `update` method → error (reset/restore, then deploy).
- Abandoned Windows mutex is not treated as “node running”.
- `Neo.BuildTasks` stamp lives under `IntermediateOutputPath` and is deleted on Clean/Rebuild.

### How to test
```shell
dotnet test test/test.workflowvalidation/test.workflowvalidation.csproj
dotnet test test/test-build-tasks/test-build-tasks.csproj
dotnet build src/neoxp/neoxp.csproj
# with a running chain:
neoxp contract deploy path/to/Contract.nef genesis -i default.neo-express
# should HALT, not FAULT Insufficient GAS
```

Split from #835 (1 of 3).


Jim8y (changes requested):
Thanks for splitting this. I found several blockers:

- The fee pad is added by every online and offline transaction path, so transfers, invokes, votes, and other transactions also overpay 0.1 GAS. Keep it scoped to deploy/update and pass --gas through force-update.
- The running-chain BuildTasks fallback breaks quoted paths, drops deploy options, forces --force, skips non-deploy commands, and can still report success. Preserve batch semantics or fail explicitly.
- The fixed 30-second confirmation timeout can expire before a configured block interval.
- TryOpenExisting does not acquire the mutex, so it cannot detect abandonment.

Please add end-to-end coverage for deploy, force-update, delayed confirmation, and FAULT reporting.

cschuchardt88:
Addressed the review on `fix/neoxp-deploy-confirm-and-fees`:

- 0.1 GAS invokescript pad is only applied on deploy/update; transfers, invokes, and votes no longer overpay. `--gas` is forwarded through force-update (and `contract update`).
- Running-chain BuildTasks fallback re-runs original `contract deploy` lines (quoted paths and options preserved, no injected `--force`). Any non-deploy batch line fails explicitly instead of reporting success.
- Confirmation timeout is `max(30s, 3 * MillisecondsPerBlock + 5s)`.
- `TryOpenExisting` now `WaitOne(0)` so an abandoned mutex is treated as not running.
- Tests cover fee scoping, force-update gas, delayed HALT, FAULT reporting, mutex hold/abandon, and batch quoting.

Jim8y (changes requested):
Thanks, the implementation issues from my review are fixed, and I reproduced deploy, same-hash force-update, and FAULT exit behavior on a live node. Two items remain:

- Please add the requested live workflow regression. The new tests use StubExpressNode and verify argument plumbing and polling, but they do not execute deploy or update against persisted chain state, so the original invokescript-versus-persist fee regression could return unnoticed.
- Please update the deploy and update command reference for --gas, its default, and the same-hash --force update behavior. The current docs omit --gas and still describe --force only as allowing a conflicting-name deployment.

Once those land and CI remains green, this is ready for rereview.

cschuchardt88:

- Live workflow tests now deploy `registrar.nef` against a real RocksDB OfflineNode, assert HALT on the application log and that the contract is listed, then `deploy --force` the same hash through `TransactionExecutor` (persisted update).
- Command reference documents `--gas` (default 1) on deploy/update, the 0.1 GAS pad, FAULT-as-error, and same-hash `--force` calling `update`.

Jim8y (changes requested):
Two issues still block this:

- Include the VM exception message in deploy/update failures so errors such as `Insufficient GAS` are not discarded.
- After confirmation succeeds, report the transaction as confirmed rather than only submitted.

Please add focused assertions for both paths.

cschuchardt88:

- Deploy/update FAULT now includes the VM exception (`Transaction {hash} FAULT: Insufficient GAS`).
- After HALT, deploy/update print **confirmed** instead of submitted (`WriteDeployResultAsync` and `WriteTxHashAsync(..., confirmed: true)` for `contract update`).
- Tests: `EnsureTransactionSucceeded_reports_FAULT` asserts the exception text; `WriteTxHashAsync_reports_confirmed_after_success` and the live force-update workflow assert `confirmed` and not `submitted`.

Jim8y (changes requested):
The FAULT detail and confirmed output fixes are correct, and the focused tests pass. One blocker remains: batch `contract deploy` still hard-codes `1m`, and batch `contract update` has no `--gas` option either. Add `--gas|-g` to both batch models, forward it to the executor, and cover parsing plus execution so batch behavior matches the direct commands.

Jim8y (dismissed):
Batch deploy and update now forward the configured gas, with parser and workflow coverage. All required checks pass.

Jim8y (approved):
The current master merge preserves async update-data resolution and the configured update gas. Workflow validation, build-task tests, Release build, format, and all three platform checks pass.

ajara87 (approved):

shargon (approved):

PR #837, opened 2026-08-25
feat: official C# template samples and wizard starters (review 2nd)

cschuchardt88:
## Summary

Samples + wizard half of what was #835.

**Depends on** the neoxp deploy/fee PR (`fix/neoxp-deploy-confirm-and-fees`). This branch is stacked on that one; review the **latest commit** (`feat: add official C# template samples and wizard starters`) if the first commit is already in the dependency PR.

### Changes
- Official [Neo.SmartContract.Template](https://github.com/neo-project/neo-devpack-dotnet/tree/master-n3/src/Neo.SmartContract.Template) starters under `samples/examples/` (Blank, NEP-17, NEP-11, Oracle, Ownable) with `Neo.BuildTasks`.
- `dotnet build` in an example folder creates `default.neo-express` if missing, compiles, and deploys as genesis while Express is running.
- **New contract** wizard uses those official C# overlays (plus the existing Storage scaffold). Official unit-test projects are not copied.

### How to test
```shell
cd samples
dotnet tool restore
dotnet tool run neoxp -- create -f -o examples/Nep17/default.neo-express
# terminal A
dotnet tool run neoxp -- run -i examples/Nep17/default.neo-express --seconds-per-block 1
# terminal B
dotnet build examples/Nep17
dotnet tool run neoxp -- contract run -i examples/Nep17/default.neo-express Nep17Contract symbol --results
```

`dotnet rebuild examples/Nep17` should reset + redeploy (batch stamp cleared — from the CLI PR).

Wizard tests: `cd extentions/neo3-visual-tracker && npm test` (template tests).

Split from #835 (2 of 3).


cschuchardt88:
Stacked on **#836**. Please review the latest commit only until #836 merges.

Jim8y (changes requested):
Thanks for splitting this. I found three blockers in the generated workflows:

- Fresh Blank, Oracle, and Ownable projects do not build because their tests call members with the wrong casing.
- The generated test projects look for default.neo-express while they create a differently named chain file, so dotnet test fails.
- A running-chain rebuild can report success and write the deploy stamp after deployment FAULTs because the target omits --force and ignores the deploy exit code.

Please add an end-to-end test that hydrates, builds, and tests every starter, and only write the deploy stamp after confirmed success. Also replace dotnet rebuild with dotnet build -t:Rebuild.

cschuchardt88:
Addressed the review on the C# starters:

- Blank, Oracle, and Ownable tests call ABI members (`myMethod` / `getResponse`).
- Generated test projects create and use `default.neo-express` (same file `FindChain()` looks for).
- Running-chain example deploy uses `--force`, does not ignore the deploy exit code, and only stamps after success.
- Docs use `dotnet build -t:Rebuild`.
- Hydrate + `dotnet build` coverage for every official overlay starter.

Jim8y (changes requested):
Thanks, the ABI casing and generated chain-path issues are fixed; all five hydrated starters now build and their generated tests pass locally. Two blockers remain:

- The samples still pin Neo.Express 3.10.1. With a running chain, dotnet build -t:Rebuild submits the --force deployment and exits successfully, but the application log is FAULT: Insufficient GAS, and the deploy stamp is still written. Stacking #836 source does not change the local tool version used by the samples. Please make this workflow use the fixed CLI and cover the actual application-log result.
- On a clean checkout, the README command dotnet test samples/test fails with MSB4057 because GetNeoContractInfo is unavailable.

Please also make the hydration test build or run the generated test projects, rather than only src/*.csproj.

cschuchardt88:

- Sample `Directory.Build.props` builds and execs this repo’s `neoxp` (not NuGet 3.10.1). That CLI waits for the application log and treats FAULT as failure, so the deploy stamp is not written on Insufficient GAS. `Configuration` defaults to `Debug` so the DLL path is never `bin//net10.0`.
- `samples/test` restores the referenced contract project before `GetNeoContractInfo`, so `dotnet test samples/test` works on a clean checkout.
- The hydration test now `dotnet build`s each generated test project (batch skipped so CI does not need a live chain).

cschuchardt88:
Restacked on #836 (`cfcb4e1`, signed). This branch now includes the FAULT VM exception text and deploy **confirmed** output.

Jim8y (approved):

ajara87 (dismissed):

samples/src/checkpoints/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
shargon, line 2: if the gitignore is to preserver the folder, is better to create the folder when not exists

extentions/neo3-visual-tracker/resources/new-contract/csharp-starters/blank/src/$_CLASSNAME_$.cs.template.txt
```diff
@@ -0,0 +1,83 @@
+using Neo.SmartContract.Framework;
+using Neo.SmartContract.Framework.Attributes;
+using Neo.SmartContract.Framework.Native;
+using Neo.SmartContract.Framework.Services;
+using System;
+using System.ComponentModel;
+
+namespace $_CONTRACTNAME_$
```
shargon, line 8: can we clone the samples from dev-pack? multiple places to mantain the same data

Jim8y (commented):
Please address Shargon's two unresolved review points before this can be considered complete: create the ignored checkpoint folder when it is absent, and clarify whether the samples should be sourced from neo-devpack-dotnet to avoid duplicated maintenance.

Jim8y (changes requested):
The current head still has two unresolved Shargon review points. The checkpoint .gitignore is present, but the new-directory path is not covered by a focused test; please add that behavioral coverage. The C# starter sources are duplicated under neo-express and neo-devpack-dotnet, so please document or implement a single source of truth before merge. I am not approving while these remain unresolved.

samples/test/contract-test.csproj
```diff
@@ -9,13 +10,16 @@
   <ItemGroup>
     <NeoContractReference Include="..\src\contract.csproj" />
   </ItemGroup>
+  <Target Name="RestoreNeoContractReferences" BeforeTargets="PrepareNeoContractGeneration">
```
cschuchardt88: `RestoreNeoContractReferences` is already defined in `Neo.BuildTasks.targets` (imported from `nuget.g.targets` after this file). The later definition keeps the `ProjectReference` injection and drops this `MSBuild Restore` body, so the clean-checkout `GetNeoContractInfo` / MSB4057 fix never runs.

Give this a unique name and hook it `BeforeTargets="PrepareNeoContractReferences"`.

samples/examples/Directory.Build.props
```diff
@@ -0,0 +1,63 @@
+<Project>
+  <PropertyGroup>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <TargetFramework>net10.0</TargetFramework>
+    <Configuration Condition="'$(Configuration)'==''">Debug</Configuration>
+    <NeoTestVersion>3.10.1</NeoTestVersion>
+    <NeoExpressInputFile>default.neo-express</NeoExpressInputFile>
+    <NeoExpressDeployStamp>$(IntermediateOutputPath)$(NeoContractName).neoxp.deploy.touch</NeoExpressDeployStamp>
```
cschuchardt88: `Directory.Build.props` is imported before the SDK sets `IntermediateOutputPath`, so this expands to `$(NeoContractName).neoxp.deploy.touch` in the project directory (same empty-property pitfall as `bin//net10.0` before `Configuration` was defaulted).

Set the stamp inside the deploy/clean targets (or under `obj\$(Configuration)\`) so it actually lives in `obj/` and is gitignored.

samples/invoke-files/TestContract.neo-invoke.json
```diff
@@ -0,0 +1,6 @@
+[
+  {
+    "contract": "#TestContract",
```
cschuchardt88: The sample contract was renamed to `SampleContract` (`NeoContractName`, `.nef`, tests). `#TestContract` will not resolve. Same for `contract.neo-invoke.json` and `contract-1.neo-invoke.json`, which still use `#contract`.

samples/invoke-files/contract.neo-invoke.json
```diff
@@ -0,0 +1,6 @@
+[
+  {
+    "contract": "#contract",
```
cschuchardt88: `#contract` no longer matches the sample. Use `#SampleContract` (and drop the extra copies unless they invoke different operations).

samples/invoke-files/contract-1.neo-invoke.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "contract": "#contract",
```
cschuchardt88: Same rename miss as the other invoke files: `#contract` should be `#SampleContract`.

samples/examples/README.md
```diff
@@ -0,0 +1,69 @@
...
+[Neo.SmartContract.Template](https://github.com/neo-project/neo-devpack-dotnet/tree/master-n3/src/Neo.SmartContract.Template)
+starters, laid out for Neo Express and `Neo.BuildTasks`. Official unit-test projects are not
+included.
+
+Walkthrough: [docs/getting-started.md](../../docs/getting-started.md).
+
+## Templates
+
+| Folder | `dotnet new` short name | Contract |
+| ------ | ----------------------- | -------- |
```
cschuchardt88, line 13: `master-n3` of neo-devpack-dotnet has `neocontractnep17` / `neocontractnep11` / `neocontractoracle` / `neocontractowner` / `neocontractsolution`, not `neocontract`. Blank here is the solution template's `Contract`, not a `dotnet new neocontract` short name.

samples/examples/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Layout
+
+```
+*.csproj              # `dotnet build` here compiles and deploys
+src/*.cs
+express.batch         # optional offline reset+deploy (`neoxp batch`)
+default.neo-express   # created on first `dotnet build` if missing
+```
+
```
cschuchardt88, line 28: `Directory.Build.props` never sets `NeoExpressBatchFile`. `dotnet build` deploys via `DeployContractToNeoExpress`; `express.batch` is only the optional offline path.

samples/examples/README.md
```diff
@@ -0,0 +1,69 @@
...
+Shared `Directory.Build.props` sets `Neo.SmartContract.Framework`, `Neo.BuildTasks`,
+`NeoExpressBatchFile`, and creates the chain file on first build.
+
+Local tools (`neoxp`, `nccs`) come from [`samples/.config/dotnet-tools.json`](../.config/dotnet-tools.json).
+
```
cschuchardt88, line 34: This heading is empty; the next line is "From an example folder:".

samples/test/contract-tests.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace ContractTests
 {
-    [CheckpointPath("checkpoints/contract-deployed.neoxp-checkpoint")]
+    [CheckpointPath("src/checkpoints/contract-deployed.neoxp-checkpoint")]
```
cschuchardt88: Batch create is CWD-relative (`samples/src/checkpoints/...` when the contract project runs the batch), which matches this path. `samples/checkpoints/contract-deployed.neoxp-checkpoint` is still committed and unused — remove it or the fixture can pick the stale file if a path walks through `samples/`.

extentions/neo3-visual-tracker/src/extension/templates/samples.examples.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    "utf8"
+  );
+  assert.match(props, /Neo\.BuildTasks/);
+  assert.match(props, /DeployContractToNeoExpress/);
+  assert.match(props, /contract deploy/);
+  assert.match(props, /NeoExpressCli/);
+  assert.match(props, /Configuration Condition="'\$\(Configuration\)'==''">Debug/);
+  assert.match(props, /src\\neoxp\\neoxp\.csproj|src\/neoxp\/neoxp\.csproj/);
+  assert.match(props, /_NeoDeployForce/);
+  assert.match(props, / --force/);
```
cschuchardt88, line 31: This only asserts files and props text. Hydration builds the wizard overlays, not `samples/examples/*` with this `Directory.Build.props`, so the repo-neoxp deploy/stamp path is untested in CI. A `dotnet build` of each example csproj (batch/deploy skipped if a live chain is too heavy) would have caught the empty `IntermediateOutputPath` stamp.

cschuchardt88 (commented):
## Summary

Official C# starters (Blank, NEP-17, NEP-11, Oracle, Ownable) land under `samples/examples/` with a custom `Directory.Build.props` that builds repo `neoxp` and deploys on `dotnet build`, and the same overlays are offered in the New contract wizard. Earlier blockers (ABI casing, `default.neo-express`, stamp-after-FAULT, hydration `dotnet build`) look addressed. Shargon's two threads (create the checkpoint directory when missing; one source of truth vs neo-devpack-dotnet) are still open and are not re-filed here.

New problems: the clean-checkout Restore target is overwritten by `Neo.BuildTasks`, the example deploy stamp is evaluated before `IntermediateOutputPath` exists, and the new invoke files still name `TestContract`/`contract` after the sample was renamed to `SampleContract`.

## Issues

### bug

- `samples/test/contract-test.csproj:13` — `RestoreNeoContractReferences` reuses the `Neo.BuildTasks` target name, so the `Restore` body never runs. Status: open
- `samples/examples/Directory.Build.props:9` — `$(IntermediateOutputPath)` is empty in `Directory.Build.props`; the deploy stamp is written next to the csproj. Status: open
- `samples/invoke-files/TestContract.neo-invoke.json:3` — invoke files still target `#TestContract` / `#contract` after the rename to `SampleContract`. Status: open
- `samples/invoke-files/contract.neo-invoke.json:3` — same `#contract` mismatch. Status: open
- `samples/invoke-files/contract-1.neo-invoke.json:3` — same `#contract` mismatch. Status: open

### suggestion

- `samples/examples/README.md:13` — `dotnet new` short name `neocontract` is not a template on neo-devpack-dotnet `master-n3`. Status: open
- `samples/examples/README.md:28` — docs say `Directory.Build.props` sets `NeoExpressBatchFile`; the examples use a custom deploy target instead. Status: open
- `samples/test/contract-tests.cs:13` — tests now load `src/checkpoints/...`, but `samples/checkpoints/contract-deployed.neoxp-checkpoint` is still committed. Status: open
- `extentions/neo3-visual-tracker/src/extension/templates/samples.examples.test.ts:31` — layout-only; CI never `dotnet build`s the example projects that use the custom deploy targets. Status: open

### nit

- `samples/examples/README.md:33` — leftover empty "From the repository root:" heading. Status: open

Jim8y (approved):
Verified the nested checkpoint-directory test, sample paths, restore target, deploy stamp, and source-of-truth documentation. Required CI is green.

shargon (commented):
I think examples should be moved to templates in neo-devpack, otherwise we will mantain multiple places for the same

Jim8y:
Shargon raised a source-of-truth concern: keeping the C# examples here duplicates the DevPack templates. Please resolve that ownership decision before merge so future template changes cannot drift.

PR #818, opened 2026-07-16
Redesign the VS Code Contract Studio workflow

Jim8y:
## Summary

- redesign the VS Code invocation editor as Contract Studio while keeping contract source visible beside the workflow
- make the selected network, signing account, and witness scope explicit before execution
- add contract-aware method and argument inputs, reusable invocation steps, JSON editing, and responsive narrow-panel behavior
- track submitted transactions from pending state through confirmation and expose their receipts
- add an **Open Contract Studio** command to workspace contract and NEF entry points
- correctly index the full contract state returned by Neo Express so ABI metadata remains available after connecting
- improve keyboard navigation, dialog behavior, focus states, labels, and extension documentation

## Evidence

The screenshot below is from a real Extension Development Host connected to a local Neo Express instance. It shows the contract source, connected blockchain, selected signer and witness scope, ABI-derived argument input, and a confirmed invocation in one workspace.

Contract Studio running a confirmed invocation

The confirmed invocation completed with VM state `HALT`. Transaction:

`0xb726c5b63127ea06362c9dec106cc3508f0314baf05e096ca7bf6a351e85815f`

The contract state was read back through Neo Express and returned `42`, matching the value submitted through Contract Studio.

## Validation

- `npm run typecheck`
- `npm test` - 43 tests passed
- `npm run lint` - 0 errors
- `npm run compile-prod`
- packaged the extension as a VSIX with a locally built Neo Express 3.10.1 package
- exercised connection, account and witness selection, invocation, confirmation, receipt inspection, JSON editing, step management, saving, and wide/narrow layouts in an Extension Development Host

## Merge order

Merge after #723 so the `neo-contract` live launch path is available.

PR #838, opened 2026-08-25
fix(devtracker): make Visual DevTracker usable again in VS Code (review 3rd)

cschuchardt88:
## Summary

VS Code DevTracker UX half of what was #835.

**Depends on** the samples/wizard PR (`feat/official-csharp-samples`) and the neoxp CLI PR. This branch is the top of the stack; review the **latest commit** (`fix(devtracker): make Visual DevTracker usable again in VS Code`).

### Changes
- Quick Start is a **tree view** (main sidebar scrollbar). **Deploy a contract** stays; **Redeploy** is removed.
- Contract Studio: dropdowns, scroll on short panes, persistent results, Output **Neo Express Invocation**.
- Wallets: genesis + node1 + NEP-6. Reset stops leftover processes (including after F5) without a hung status bar.
- F5 prefers `src/neoxp/bin/.../neoxp.dll` and only passes `-g` when `contract deploy --help` lists `--gas`.
- Getting-started docs and an F5 launch config.

### How to test (F5)
```shell
dotnet build src/neoxp/neoxp.csproj
cd extentions/neo3-visual-tracker
npm install
npm test
npm run compile
```

Open `extentions/neo3-visual-tracker` in VS Code → **Launch Visual DevTracker**. In the Extension Development Host, open a **fresh folder**.

1. Quick Start first, sidebar scrollbar.
2. Create/start Express → **genesis** and **node1** in Wallets.
3. New contract → C# Ownable → Deploy as genesis.
4. Rocket → Contract Studio → shrink the pane, scroll to **Run invocation** → `getOwner` result stays visible.
5. Reset Neo Express; status bar must not stick on “Restarting node…”. Deploy again.

Split from #835 (3 of 3).


cschuchardt88:
Stacked on **#836** and **#837**. Please review the latest commit only until those merge.

cschuchardt88:
Markdown (getting-started, debugger reference, README/doc pointers) moved to a dedicated PR so this one stays on DevTracker UX.

Jim8y (changes requested):
Thanks for splitting this. I found several blockers in the DevTracker commit:

- Encrypted-wallet passwords are passed in the options array while the runner logs full slow and timed-out commands, exposing --password values.
- PTY handlers are registered after terminal.show(), so fast startup output can be missed and produce a false timeout.
- Ordinary workspace-contract selection again creates a new invocation file, reversing #818.
- Commands for a non-active Express item can use wallets from the active chain.
- Quick Start hides transfers in a fresh Express workspace because Express wallets do not satisfy hasWallets.

Please add focused behavioral tests for these paths before rereview.

cschuchardt88:
Addressed the DevTracker review:

- `--password` values are redacted in slow/timeout CLI logs.
- PTY start watchers are registered before `createTerminal`/`show()` so fast "Neo express is running" output is not missed.
- Selecting a workspace contract reuses `invoke-files/<name>.neo-invoke.json` instead of creating `<name>-1`.
- Transfer/deploy account lists come from the selected Express identifier, not the active chain's autocomplete wallets.
- Quick Start shows transfer actions for a fresh Express instance even when there are no NEP-6 wallets.
- Focused tests cover those paths.

Jim8y (changes requested):
Thanks. Password redaction, PTY watcher ordering, and fresh-Express transfer actions are fixed.

Two blockers remain:

- Workspace contract nodes still open Contract Studio on ordinary selection and create the canonical invoke file on the first click. Reusing that file only avoids the numeric suffix; it does not restore the #818 behavior. Workspace nodes should have no default command and keep the explicit rocket action.
- Wallet choices now come from the selected Express config, but ensureAccountHasGas still resolves a matching name through the active connection's autocomplete data. With Express A active and a command launched for Express B, balance checks and auto-funding can still target A's same-named address on B. Please resolve the address from the selected identifier and cover this scenario behaviorally.

Please also confirm whether removing workspace NEP-6 signers from Express deploy/transfer is intentional.

cschuchardt88:

- Workspace contract nodes no longer have a default command. Ordinary selection does not open Contract Studio or create an invoke file. The rocket inline action is unchanged (`viewItem == workspaceContract`).
- `ensureAccountHasGas` resolves the address from the **selected** Express identifier (`getWalletAddresses()`), not the active connection’s autocomplete. Covered with `resolveAccountForIdentifier` tests (Express A active / command for Express B).

Removing workspace NEP-6 signers from Express deploy/transfer was **not** intentional. They are restored: account lists are selected-chain Express wallets **plus** workspace NEP-6 files (`accountSigners[name] !== name`). Active-chain Express names are still excluded so they cannot leak onto another instance.

Jim8y (changes requested):
Thanks. The workspace default-command and selected-chain GAS fixes are correct, and NEP-6 signers are restored. One collision case still blocks rereview: if the selected Express config contains wallet owner and the workspace also has owner.json, the picker deduplicates them to owner, the transaction uses the NEP-6 file, but the GAS check and funding use the Express address. Please disambiguate the two choices or make one identity consistently take precedence, and add a focused collision test.

cschuchardt88:
Addressed the owner collision:

If the selected Express instance has `owner` and the workspace has `owner.json`, the picker now shows both:

- `owner` — Express wallet (GAS check and tx use the Express address)
- `owner (owner.json)` — workspace NEP-6 file (GAS check and tx use the file)

Covered by `buildAccountChoices disambiguates Express owner from workspace owner.json`.

cschuchardt88:
Restacked on #837 / #836 (signed). Includes the FAULT exception message and deploy **confirmed** output from #836.

ajara87:
Conflicts

Jim8y (commented):
The behavioral fixes are covered by the follow-up notes, but this branch is currently conflicting with `master`. Please rebase and rerun the focused extension tests before requesting rereview.

Jim8y (changes requested):
Two account cases still need fixing. Selecting a nondefault NEP-6 account checks/funds its address, but passes only the wallet path, so neoxp signs with the default account. Confirmed with real wallets. Also, same-named wallets in different folders collapse into one choice. Please preserve account identity end to end, or expose only the supported default account, and add behavioral tests.

Jim8y (changes requested):
The latest head still needs the account-identity fixes from the current Jim8y review. A non-default NEP-6 account can be balance-checked by its address but is passed to neoxp by wallet path, which may sign with the wallet default account; same-named wallets from different folders can also collapse into one choice. Please preserve the selected account identity through signing and add behavioral coverage before merge.

extentions/neo3-visual-tracker/src/shared/expressWalletAddresses.ts
```diff
@@ -0,0 +1,182 @@
...
+    address: expressWallets[name],
+  }));
+  for (const name of sortWalletNames(workspaceNep6AccountNames(accountSigners))) {
+    const path = accountSigners![name];
+    const label = expressSet.has(name)
+      ? `${name} (${walletFileName(path)})`
+      : name;
+    choices.push({
+      label,
+      signer: path,
```
cschuchardt88, line 196: **bug** — NEP-6 signer is the wallet path, not the selected account.

`address` is the chosen account, but deploy/transfer/invoke pass `signer` (this path) to neoxp. `TryGetNEP6Wallet` only loads `IsDefault` or the single account, so a non-default NEP-6 account can be GAS-checked/funded and then signed by the default. Contract Studio does the same (`accountSigners[name] = wallet.path`).

Keep the selected account through signing, or expose only the default/single account neoxp supports. Add a behavioral test with a two-account wallet.

extentions/neo3-visual-tracker/src/extension/autoComplete.ts
```diff
@@ -174,19 +186,32 @@ export default class AutoComplete {
...
         newData.addressNames[account.address] =
           newData.addressNames[account.address] || [];
-        newData.addressNames[account.address].push(
-          account.label || wallet.path
-        );
+        newData.addressNames[account.address].push(displayName);
         newData.addressNames[account.address] = dedupeAndSort(
           newData.addressNames[account.address]
         );
+        if (!newData.wellKnownAddresses[displayName]) {
```
cschuchardt88, line 216: **bug** — Same-named workspace wallets collapse to one picker choice.

First `displayName` wins (`alice.json` in two folders, or two accounts labeled `Treasury`). Express collisions append only the basename, so different directories still collide. The Wallets tree lists both; `accountSigners` / `buildAccountChoices` keep one.

Key by path (and account), not basename. Cover with a focused collision test.

cschuchardt88 (commented):
## Summary

DevTracker UX is in better shape on this stacked PR: Quick Start is a sidebar tree, Contract Studio has real dropdowns/scroll/persistent results, passwords are redacted, PTY start watchers register before `show()`, workspace contracts have no default command, deploy/transfer wallets come from the selected Express config, and `owner` vs `owner.json` is labeled. CI is green and the branch is mergeable.

Two account-identity bugs from the current Jim8y review are still present on `318a98c`. `Keep selected workspace account identity` split GAS lookup from the signer string, but neoxp still signs a NEP-6 file by wallet path (default account only), and same-named workspace wallets still collapse. `neoExpressCommands.accounts.test.ts` only greps source; it does not cover those paths.

## Issues

### bug — File `extentions/neo3-visual-tracker/src/shared/expressWalletAddresses.ts:163` — Status: open

Selecting a non-default NEP-6 account checks and funds `address`, then deploy/transfer/invoke pass `signer: path` to neoxp. `TryGetNEP6Wallet` loads `IsDefault` or the only account, so the default account signs. Contract Studio `runFile` does the same via `accountSigners[name] = wallet.path`. Preserve the selected account through signing, or list only the default/single account neoxp supports. Add a behavioral test with a two-account wallet.

### bug — File `extentions/neo3-visual-tracker/src/extension/autoComplete.ts:211` — Status: open

`if (!newData.wellKnownAddresses[displayName])` drops the second wallet when two files share a display name (`wallets/a/alice.json` and `wallets/b/alice.json` both become `alice`; two `Treasury` labels likewise). Express collisions append only the basename, so different directories still collide. The Wallets tree lists both; the picker keeps one. Key by path (and account). Cover with a focused collision test.

Jim8y (approved):
Verified the supported-account filtering, unique wallet labels, focused tests, and full extension CI. Required CI is green.

PR #839, opened 2026-08-25
docs: getting-started guide and CLI/VS Code walkthroughs (review 4th)

cschuchardt88:
## Summary

Markdown that lived in #838, split out so the DevTracker UX PR is code-only.

- New [docs/getting-started.md](docs/getting-started.md) (CLI and VS Code).
- New [docs/debugger-command-reference.md](docs/debugger-command-reference.md).
- Pointers from existing docs (`quickstart`, `installation`, command references) and the repo / DevTracker READMEs.

**Depends on #837** (samples paths in the getting-started guide). This branch is stacked on `feat/official-csharp-samples`; review the **latest commit** until #837 merges.

## How to test

Read [docs/getting-started.md](docs/getting-started.md) end to end and confirm links resolve (`samples/examples`, DevTracker README, debugger reference).

PR #841, opened 2026-08-29
fix: reject invalid hashes in show transaction

cschuchardt88:
## Summary

`neoxp show transaction` used `UInt256.Parse`, which throws `FormatException` on garbage input.

Use `TryParse` and report `Invalid transaction hash \"...\"`.

Independent of #836–#839.


Jim8y (approved):

ajara87 (approved):

shargon (commented):
Lgtm

PR #842, opened 2026-08-29
fix: parse contract update data with the chain contract resolver

cschuchardt88:
## Summary

`contract update --data` and batch `contract update` used `TransactionExecutor.ContractParameterParser`, which only resolves account names — not deployed contract names.

Use `GetContractParameterParserAsync` so `--data` JSON can refer to contracts by name, matching invoke/run.

Independent of #836–#839.

### DIFF
--- a/src/neoxp/Commands/BatchCommand.cs
+++ b/src/neoxp/Commands/BatchCommand.cs
@@ -244,8 +244,10 @@ namespace NeoExpress.Commands
                             }
                         case CommandLineApplication<BatchFileCommands.Contract.Update> cmd:
                             {
-                                var parser = await txExec.ExpressNode.GetContractParameterParserAsync(chainManager.Chain).ConfigureAwait(false);
-                                var data = ContractCommand.Update.ParseUpdateData(cmd.Model.Data, s => parser.ParseParameter(s));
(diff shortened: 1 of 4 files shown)


Jim8y (dismissed):

src/neoxp/Commands/BatchCommand.cs
```diff
@@ -244,7 +244,8 @@ await txExec.ContractInvokeAsync(
                             }
                         case CommandLineApplication<BatchFileCommands.Contract.Update> cmd:
                             {
-                                var data = ContractCommand.Update.ParseUpdateData(cmd.Model.Data, txExec.ContractParameterParser);
+                                var parser = await txExec.ExpressNode.GetContractParameterParserAsync(chainManager.Chain).ConfigureAwait(false);
+                                var data = ContractCommand.Update.ParseUpdateData(cmd.Model.Data, s => parser.ParseParameter(s));
```
ajara87: add this two lines in a helper?

ajara87:
could you add a test for contract nsme resolver?

Jim8y (approved):
The shared parser now covers both command paths and skips contract lookup when update data is omitted. Focused tests and all platform checks pass.

ajara87 (approved):

shargon (approved):

PR #844, opened 2026-08-29
fix: show transaction for unconfirmed online txs

cschuchardt88:
## Summary

On a running node, `show tx` always called `getapplicationlog`. That RPC throws `-100 Unknown transaction/blockhash` for mempool txs, so unconfirmed hashes failed even though `getrawtransaction` succeeded.

Catch that error and return `(tx, null)`, matching `OfflineNode.GetTransaction`.

Independent of #836–#839.

### DIFF
--- a/src/neoxp/Node/OnlineNode.cs
+++ b/src/neoxp/Node/OnlineNode.cs
@@ -242,9 +242,21 @@ namespace NeoExpress.Node
         {
             var hash = txHash.ToString();
             var response = await rpcClient.GetRawTransactionAsync(hash).ConfigureAwait(false);
-            var log = await rpcClient.GetApplicationLogAsync(hash).ConfigureAwait(false);
+            RpcApplicationLog? log = null;
+            try
+            {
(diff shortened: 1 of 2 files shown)


shargon (commented):
Is not better to fix GetApplicationLog?

cschuchardt88:
`GetApplicationLog` throwing `-100` is the right RPC behavior. A mempool hash has no persisted execution log, and the same error is used for a completely unknown hash. Changing that RPC to return null would hide real misses from every other client (`trace`, neo-cli, dApps).

`IExpressNode.GetTransactionAsync` is already `(tx, log?)`. `OfflineNode` returns `null` for the log when `GetAppLog` misses. This catch only maps that same `-100` at the `OnlineNode` boundary so `show tx` can print an unconfirmed transaction the way offline mode already does.

If we still want a helper, it belongs on `RpcClient.GetApplicationLogAsync` as `throwOnError: false` (or a Try* method), not on the `getapplicationlog` RPC itself.

Jim8y (approved):

ajara87 (approved):

shargon (approved):

PR #849, opened 2026-09-18
fix: support legacy express wallet accounts

Jim8y:
## Problem
Neo Express could not load older wallet entries that only contain a private key. `wallet list` failed while parsing an empty script-hash, even though the same account format is accepted by the toolkit wallet parser.

## Changes
- derive the signature contract and script-hash when an account has a private key but omits derived fields
- preserve explicit script-hashes and support keyless accounts without dereferencing a missing contract
- resolve wallet names using the chain's address version
- add regression coverage for legacy, keyless, and wallet-name account resolution

## Validation
- `dotnet build neo-express.sln --configuration Release --no-restore`
- `dotnet test neo-express.sln --configuration Release --no-build --verbosity minimal --filter "Category!=Integration"` (781 passed)
- verified `wallet list` and `wallet export` against `test/test.bctklib/_testFiles/default.neo-express.json`


ajara87 (approved):

src/neoxp/Extensions/ExpressChainExtensions.cs
```diff
@@ -84,13 +84,18 @@ public static string ResolvePassword(this ExpressChain chain, string name, strin
             return McMaster.Extensions.CommandLineUtils.Prompt.GetPassword($"enter password for {name}");
         }
 
-        public static UInt160 GetScriptHash(this ExpressWalletAccount? @this)
+        public static UInt160 GetScriptHash(this ExpressWalletAccount? @this, byte addressVersion)
         {
             ArgumentNullException.ThrowIfNull(@this);
 
+            if (!string.IsNullOrWhiteSpace(@this.ScriptHash))
```
cschuchardt88, line 91: **suggestion:** `GetScriptHash` and `FromExpressWalletAccount` no longer share a fallback order.

This helper returns an explicit `script-hash` when present, otherwise a signature contract derived from the private key. It never consults `Contract.Script`. `DevWalletAccount.FromExpressWalletAccount` does the opposite for the omitted-hash case: stored contract script → signature contract from the key.

An account with a custom redeem script, empty `script-hash`, and a private key therefore resolves to two different hashes depending on which API is used (single-sig vs the stored contract). Name resolution (`TryGetAccountHash`, `ParseScriptHashToBlockAsync`) takes this path.

Consider one shared helper: explicit address → `Contract.Script` hash → signature contract from the private key → `FormatException`.

src/neoxp/Extensions/ExpressChainExtensions.cs
```diff
@@ -101,7 +106,7 @@ public static bool TryGetAccountHash(this ExpressChain chain, string name, [Mayb
                 {
                     if (string.Equals(name, chain.Wallets[i].Name, StringComparison.OrdinalIgnoreCase))
                     {
-                        accountHash = chain.Wallets[i].DefaultAccount.GetScriptHash();
+                        accountHash = chain.Wallets[i].DefaultAccount.GetScriptHash(chain.AddressVersion);
```
cschuchardt88: **suggestion:** `TryGetAccountHash` still assumes `DefaultAccount` is non-null.

`ExpressWallet.DefaultAccount` is `Accounts.SingleOrDefault(a => a.IsDefault)`. The in-repo legacy fixture `test/test.bctklib/_testFiles/default.neo-express.json` wallet `mintest` is private-key-only and omits `is-default`, so `DefaultAccount` is null. This `Try*` method then throws `ArgumentNullException` from `GetScriptHash` instead of returning false.

`Wallet.GetDefaultAccount()` falls back to the first account; `ToolkitWalletTests.ParseWallet("mintest")` relies on that. Worth using `DefaultAccount ?? Accounts.FirstOrDefault()` here (and on the consensus-node path below), or covering that fixture in `AccountNameResolutionSupportsLegacyAndKeylessAccounts`.

test/test.workflowvalidation/DevWalletTests.cs
```diff
@@ -27,4 +34,76 @@ public void WalletRemainsUnlockedAndRenamable()
...
+        var keyPair = new KeyPair(Convert.FromHexString(PrivateKey));
+        var contract = Contract.CreateSignatureContract(keyPair.PublicKey);
+
+        Assert.Equal(contract.ScriptHash, account.ScriptHash);
+        Assert.Equal(contract.Script, account.Contract!.Script);
+        Assert.Equal(keyPair.PrivateKey, account.GetKey()!.PrivateKey);
+    }
+
+    [Fact]
+    public void KeylessAccountRoundTripsWithoutInventingAContract()
```
cschuchardt88, line 72: **suggestion:** Keyless accounts now parse, but the command that motivated this PR still cannot list them.

`WalletCommand.List.PrintAccountInfo` does `account.GetKey() ?? throw new Exception()` (and export calls `NEP6Wallet.CreateAccount(account.Contract, account.GetKey())`, which NREs when `Contract` is null). After this change, a watch-only account gets past `FromExpressWalletAccount` and then fails with an empty `Exception` instead of a parse error.

If keyless support is in scope, list/export need a watch-only branch. If this is only defensive parsing, this test oversells the feature — `wallet list` of a file that mixes legacy and watch-only accounts will still exit 1.

test/test.workflowvalidation/DevWalletTests.cs
```diff
@@ -27,4 +34,76 @@ public void WalletRemainsUnlockedAndRenamable()
...
+        Assert.False(account.HasKey);
+        Assert.Null(account.Contract);
+        Assert.Equal(source.ScriptHash, roundTrip.ScriptHash);
+        Assert.Equal(source.Label, roundTrip.Label);
+        Assert.Null(roundTrip.Contract);
+        Assert.Empty(roundTrip.PrivateKey);
+    }
+
+    [Fact]
+    public void AccountNameResolutionSupportsLegacyAndKeylessAccounts()
```
cschuchardt88, line 119: **nit:** `GetScriptHash` now requires `addressVersion`, but this test (and the others) only use `ProtocolSettings.Default.AddressVersion`.

A chain with a non-default `address-version` is the case that made the new parameter necessary: an explicit `script-hash` is a base58 address and `ToScriptHash` rejects a version mismatch. A regression that passed `ProtocolSettings.Default.AddressVersion` instead of `chain.AddressVersion` would still pass this test.

cschuchardt88 (commented):
## Summary

This correctly unblocks the reported `wallet list` failure on private-key-only express accounts (empty `script-hash`): `FromExpressWalletAccount` now derives the signature contract/hash like `ToolkitWallet.Account.Parse`, and `GetScriptHash` no longer requires a stored address. Passing `chain.AddressVersion` is required once the stored `script-hash` (actually a base58 address) is honored. CI is green and the private-key-only unit test matches `default.neo-express.json`'s `mintest`/`mintest2` keys.

The new keyless path and the two hash helpers are not fully aligned with each other or with CLI commands that still assume a key.

## Issues

1. **suggestion** — `src/neoxp/Extensions/ExpressChainExtensions.cs:91`
   Status: open
   `GetScriptHash` ignores `Contract.Script` and can disagree with `FromExpressWalletAccount` when a custom contract is present but `script-hash` is omitted.

2. **suggestion** — `src/neoxp/Extensions/ExpressChainExtensions.cs:109`
   Status: open
   `TryGetAccountHash` throws on wallets with no `is-default` account (in-repo `mintest` fixture).

3. **suggestion** — `test/test.workflowvalidation/DevWalletTests.cs:54`
   Status: open
   Keyless accounts parse/round-trip, but `wallet list` / `wallet export` still require a private key and contract.

4. **nit** — `test/test.workflowvalidation/DevWalletTests.cs:75`
   Status: open
   Address-version wiring is untested with a non-default `AddressVersion`.

src/neoxp/Extensions/ExpressChainExtensions.cs
```diff
@@ -84,13 +84,18 @@ public static string ResolvePassword(this ExpressChain chain, string name, strin
...
+            if (!string.IsNullOrWhiteSpace(@this.ScriptHash))
+                return @this.ScriptHash.ToScriptHash(addressVersion);
+
+            if (string.IsNullOrWhiteSpace(@this.PrivateKey))
+                throw new FormatException("Wallet account must specify a script hash or private key");
+
             var keyPair = new KeyPair(@this.PrivateKey.HexToBytes());
-            var contract = Neo.SmartContract.Contract.CreateSignatureContract(keyPair.PublicKey);
-            return contract.ScriptHash;
+            return Neo.SmartContract.Contract.CreateSignatureContract(keyPair.PublicKey).ScriptHash;
```
shargon, line 107: here is not used the `addressVersion`

Jim8y:
Addressed the remaining compatibility points:

- derive the hash from `contract.script` when `script-hash` is absent;
- resolve the first account when a legacy wallet has no `is-default` account;
- allow watch-only accounts in `wallet list` and NEP-6 export;
- added custom-contract, no-default, watch-only export, and non-default address-version coverage.

Local `DevWalletTests` passes (8/8).

Wi1l-B0t (approved):

ajara87 (approved):

shargon (approved):

PR #852, opened 2026-09-19
fix(worknet): stop the node on the RPC port it was started with

Jim8y:
## Problem

`neo-worknet stop` connected to a hardcoded RPC port (30332). A node started with a non-default `--rpc-port` could not be stopped: `stop` reported `node was not running` while the node kept listening on its actual port. The stale comment above the constant claimed the port was hardcoded in `RunCommand.GetRpcServerSettings`, which has not been true since `--rpc-port` was introduced.

## Fix

- `RunCommand` persists the effective RPC port into the worknet file (whose `consensus-nodes` entry already carries an `rpc-port` property) before the RPC server starts.
- `WorknetFile` gains an `RpcPort` property; `LoadWorknetAsync` parses it and falls back to the default port for files without the property.
- `StopCommand` connects to `worknet.RpcPort` instead of the hardcoded constant; the stale constant and comment are removed.

## Testing

- `neo-worknet` unit tests (`test.worknet`, including the existing `RunCommandPortTests`)
- End-to-end: with a worknet file recording `rpc-port: 30999`, `neo-worknet stop` connects to and stops a listener on 30999; a file without the `rpc-port` property correctly falls back to 30332
- Full solution: format / Release build / tests / pack


src/worknet/Commands/RunCommand.cs
```diff
@@ -61,6 +61,16 @@ internal async Task<int> OnExecuteAsync(CommandLineApplication app, IConsole con
...
 
             var secondsPerBlock = SecondsPerBlock ?? 0;
+
+            // persist the effective RPC port before the server starts so that
+            // `neo-worknet stop` connects to the right port even when the node
+            // was launched with a non-default --rpc-port value
+            if (RpcPort != worknet.RpcPort)
+            {
+                worknet = worknet with { RpcPort = RpcPort };
+                fs.UpdateWorknetRpcPort(filename, RpcPort);
```
cschuchardt88: **bug:** The port is persisted before `RunAsync` can fail.

`OpenDb` (already-running node / RocksDB lock) and the RPC bind both happen inside `RunAsync`, after this write. A second `neo-worknet run` while a custom-port node is up, or a bind failure on a new `--rpc-port`, rewrites `consensus-nodes[0].rpc-port` anyway. `stop` then hits the new port, reports `node was not running`, and leaves the original listener up.

Write the port only after the RPC server is listening, or restore the previous value if `RunAsync` throws. Express refuses to start when the node is already running; the same check here would also close the easy failure path.

test/test.worknet/WorknetFileTests.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.IO;
+using System.IO.Abstractions;
+using Xunit;
+
+namespace test.worknet;
+
+public class WorknetFileTests
+{
+    [Fact]
+    public void update_rpc_port_adds_missing_port_property()
```
cschuchardt88, line 43: **suggestion:** This only asserts `UpdateWorknetRpcPort` can add `rpc-port` to `{"consensus-nodes":[{}]}`.

`StopCommand` never calls that helper; it uses `LoadWorknetAsync` → `worknet.RpcPort`. There is no test that a real worknet file’s `consensus-nodes[0].rpc-port` round-trips, or that a missing property falls back to `DEFAULT_RPC_PORT`. Those are the behaviors the PR description claims.

A couple of `LoadWorknetAsync` cases (present port / absent port) would lock that in without standing up a node.

src/worknet/Models/WorknetFile.cs
```diff
@@ -16,4 +16,10 @@ namespace NeoWorkNet.Models;
 public record WorknetFile(
     Uri Uri,
     BranchInfo BranchInfo,
-    Wallet ConsensusWallet);
+    Wallet ConsensusWallet)
+{
+    // RPC port of the running instance as recorded in the worknet file's
+    // consensus-nodes entry. RunCommand refreshes it on startup so that
+    // StopCommand can reach the node when a non-default --rpc-port was used.
+    public ushort RpcPort { get; init; } = Commands.RunCommand.DEFAULT_RPC_PORT;
```
cschuchardt88: **nit:** A Models type should not take its default from `Commands.RunCommand`. Put `DEFAULT_RPC_PORT` on `WorknetFile` (or a shared constant) and have `RunCommand` / `LoadWorknetAsync` reference that.

cschuchardt88 (commented):
## Summary

`neo-worknet stop` no longer talks to hardcoded 30332. `run` writes the effective RPC port onto `consensus-nodes[0].rpc-port` (the field `SaveWorknetFile` already emitted), `LoadWorknetAsync` reads it into `WorknetFile.RpcPort` with a default-port fallback, and `stop` uses that value. That is the right place to record the port, and it fixes the case where a node started with `--rpc-port` could not be stopped.

The write happens before `RunAsync` can fail. A second `run` while the node is up (RocksDB lock) or a bind failure still rewrites the file, so `stop` can miss a live listener. The new test only checks `UpdateWorknetRpcPort` on stub JSON; it never loads a worknet file or covers the fallback `stop` relies on.

## Issues

1. **bug** `src/worknet/Commands/RunCommand.cs:71` (RIGHT) — Status: open
   `UpdateWorknetRpcPort` runs before `RunAsync`. If start fails after that write, the file no longer matches whatever is actually listening.
2. **suggestion** `test/test.worknet/WorknetFileTests.cs:24` (RIGHT) — Status: open
   No test that `LoadWorknetAsync` reads `rpc-port` or falls back to 30332, which is the path `StopCommand` uses.
3. **nit** `src/worknet/Models/WorknetFile.cs:24` (RIGHT) — Status: open
   `WorknetFile` in Models takes its default from `Commands.RunCommand.DEFAULT_RPC_PORT`.

Jim8y:
Fixed the startup consistency issue and added load/fallback coverage.

- the worknet file is updated only after the node and consensus services start successfully;
- legacy files without `rpc-port` still use the shared default `30332`;
- the default is now owned by `WorknetFile`, independent of `RunCommand`.

Local `test.worknet` passes (13/13).

Jim8y:
The first CI run only failed the repository import-order formatter. I corrected `WorknetFileTests.cs`; no logic changes were needed. The local worknet suite remains green (13/13).

shargon (approved):

PR #853, opened 2026-09-19
fix: surface trace file write errors instead of failing silently

Jim8y:
## Problem

`TraceDebugStream.Write` swallowed every exception with an empty catch block. An I/O failure or MessagePack serialization error while tracing produced a truncated `.neo-trace` file while `neotrace` still reported success, silently corrupting the debugging artifact.

## Fix

- `TraceDebugStream` captures the first write error in a public `WriteError` property and stops writing once it occurs (the file is already incomplete at that point).
- `Dispose` reports the write error to stderr (`Error writing trace file: ... The trace file may be incomplete or corrupted.`) and no longer lets flush failures escape unnoticed.
- `neotrace` fails with a non-zero exit and a descriptive error when the trace for a transaction could not be written.

## Testing

- `test.bctklib` suite (389 tests)
- Full solution: format / Release build / tests / pack


src/bctklib/trace-models/TraceDebugStream.cs
```diff
@@ -45,13 +54,37 @@ public TraceDebugStream(Stream stream)
...
+                stream.Dispose();
+            }
+            catch (Exception ex)
+            {
+                writeError ??= ex;
+            }
+
+            if (writeError is { } error)
+            {
+                Console.Error.WriteLine($"Error writing trace file: {error.Message}. The trace file may be incomplete or corrupted.");
```
cschuchardt88: **suggestion** — `Dispose` writes directly to `Console.Error` from library code. That runs even when the caller already inspects `WriteError` and throws, so `neotrace` prints this line and then `Failed to write/finalize trace file...`. It also bypasses `IConsole` and shows up in unit tests.

If this is a backstop for `ExpressApplicationEngineProvider` (which still does not check `WriteError`), consider having that caller report the error instead, and keep `TraceDebugStream` side-effect free beyond capturing `WriteError`.

Status: open

test/test.bctklib/TraceDebugStreamTests.cs
```diff
@@ -0,0 +1,35 @@
...
+using Neo.BlockchainToolkit.TraceDebug;
+using System.IO;
+using Xunit;
+
+namespace test.bctklib;
+
+public class TraceDebugStreamTests
+{
+    [Fact]
+    public void FlushFailureIsExposedAfterDispose()
```
cschuchardt88: **suggestion** — This only covers `Flush` failing in `Dispose`. The original silent failure was the empty `catch` in `Write()`. Please add a case where `Write(ReadOnlySpan<byte>)` throws: `WriteError` should be set (and further writes skipped) without requiring `Dispose`.

Status: open

src/trace/Program.cs
```diff
@@ -128,39 +128,52 @@ static async Task TraceBlockAsync(RpcClient rpcClient, Block block, ProtocolSett
...
+                        throw new Exception($"Unexpected Application Log executions count. Expected 1, got {appLog.Executions.Count}");
+                    var execution = appLog.Executions[0];
+
+                    engine.LoadScript(tx.Script);
+                    engine.Execute();
+                    if (engine.State != execution.VMState)
+                        throw new Exception($"Unexpected script execution state. Expected {execution.VMState} got {engine.State} reason: {engine.FaultException}");
+
+                    if (traceSink?.WriteError is { } writeError)
+                        throw new Exception($"Failed to write trace file for transaction {tx.Hash}: {writeError.Message}");
```
cschuchardt88: **suggestion** — Pass `writeError` as the inner exception so the original I/O or MessagePack stack is preserved: `throw new Exception($"Failed to write trace file for transaction {tx.Hash}: {writeError.Message}", writeError);`. Same for the finalize throw on line 176.

Status: open

cschuchardt88 (commented):
## Summary

Capturing the first `TraceDebugStream` write/flush failure, stopping further writes, and making `neotrace` exit non-zero is the right fix for the empty `catch` that used to leave truncated `.neo-trace` files while reporting success. The post-`using` check for flush/dispose errors is also correct: `TraceApplicationEngine` owns the sink, so those failures only appear after the engine is disposed.

A few follow-ups would make the surface complete: `Dispose` still writes to `Console.Error` from library code (duplicate CLI output, bypasses `IConsole`, and is the only report path for `ExpressApplicationEngineProvider`), the new test covers flush-on-dispose but not the original `Write` swallow, and the `neotrace` throws drop the original exception as `InnerException`.

## Issues

1. **suggestion** — `src/bctklib/trace-models/TraceDebugStream.cs:77` — `Dispose` writes directly to `Console.Error`. That fires even when the caller already inspects `WriteError` and throws, so `neotrace` prints this line and then `Failed to write/finalize trace file...`. It also bypasses `IConsole` and runs in unit tests. Prefer letting callers report `WriteError` (including `ExpressApplicationEngineProvider`, which still has no check). Status: open
2. **suggestion** — `test/test.bctklib/TraceDebugStreamTests.cs:20` — The regression test only covers `Flush` failing in `Dispose`. The original silent failure was the empty `catch` in `Write()`. A stream whose `Write(ReadOnlySpan<byte>)` throws should set `WriteError` (and stop further writes) without needing `Dispose`. Status: open
3. **suggestion** — `src/trace/Program.cs:160` — `throw new Exception($"... {writeError.Message}")` drops the original I/O or MessagePack exception. Pass `writeError` as `innerException` here and at the finalize throw on line 176. Status: open

ajara87 (approved):

shargon (dismissed):

Jim8y:
Addressed the review points:

- removed library-side `Console.Error` output;
- added a write failure test that verifies `WriteError` and stops later writes;
- preserved the original write/finalize exception as `InnerException`.

Local `test.bctklib` passes (390/390).

Jim8y:
Rebased onto the current `master` and retained the fixes after resolving the trace stream changes from #854.

The current `test.bctklib` suite passes (390/390), and `dotnet format --verify-no-changes` passes for the project.

ajara87 (approved):

Wi1l-B0t (approved):

shargon (approved):

